PR #2, opened 2016-11-26
Atomic refcounters, fixed device release

CookiePLMonster:
It could solve issues with inconsistent device refcounters in games which expect d3d8 to be thread safe.

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -42,29 +42,34 @@ ULONG STDMETHODCALLTYPE Direct3DDevice8::AddRef()
 }
 ULONG STDMETHODCALLTYPE Direct3DDevice8::Release()
 {
-	ULONG myRef = InterlockedDecrement(&_ref);
-	if (myRef <= 2)
+	ULONG myRef = InterlockedExchange(&_ref, _ref);
+	if (myRef <= 3) // 2 from _current_rendertarget and _current_depthstencil + 1 from the caller
 	{
-		if (_current_rendertarget != nullptr)
+		// Transfer ownership of these members to prevent an infinite loop and unsigned integer underflow on surface refcounts
+		// since releasing those surfaces relleases this device too
+		auto* rt = _current_rendertarget;
+		auto* ds = _current_depthstencil;
+		_current_rendertarget = nullptr;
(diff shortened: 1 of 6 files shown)


source/d3d8to9_device.cpp
```diff
@@ -36,13 +36,14 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::QueryInterface(REFIID riid, void **pp
...
-	_ref++;
+	InterlockedIncrement( &_ref );
 
 	return _proxy->AddRef();
 }
 ULONG STDMETHODCALLTYPE Direct3DDevice8::Release()
 {
-	if (--_ref <= 2)
+	ULONG myRef = InterlockedDecrement( &_ref );
+	if (myRef <= 2)
```
CookiePLMonster: This conditional looks pretty arbitrary though, why is it 2 exactly?

crosire: Because the device holds two object instances in `_current_rendertarget` and `_current_depthstencil` which in turn each hold a reference to the device. Because of this the reference count should always be exactly two rather than zero after the application cleaned up all its resources.

CookiePLMonster: I see. Fair enough!

crosire:
Looks good. But could you please remove the spaces between braces and arguments in the function calls to stay true to the rest of the code? =)

CookiePLMonster:
Sure, will do. I noticed the initial issue I was trying to fix by this is still happening at times though :(

"Reference count for 'IDirect3DDevice8' object 1AAFEA88 (2) is inconsistent."

crosire:
That's not necessarily an issue. It can happen if the game does not release its objects in the reverse order it created them, which is quite common. I don't think a lot of D3D8 games do multi-threading actually (though the API seems to support it partially). Still these changes can't hurt.

CookiePLMonster:
These interlocked ops are literally compiled to a single locked assembly instruction so they are not any slower than regular operations (the perf difference doesn't really make a difference in the context of d3d rendering) so it's a win-win.

CookiePLMonster:
I have updated the PR with a fix for device refcounts. See the comments in the code for an explaination on what happened.

crosire:
Ah. Nice catch.

PR #5, opened 2016-11-27
Fix Clang compilation

CookiePLMonster:
Compiling with v140 toolsets in /MT mode (I don't want to rely on VS2015 redist packages) gives me a RIDICULOUS amount of hits on virustotal:

https://www.virustotal.com/file/ecd4935c5a3b653062a648674c6856cbd47edd33e6f5c7e64a4277054ec261c9/analysis/1480277016/

The issue seems to be real since within 24h from the moment the update with this DLL went live I saw numerous people reporting their AVs block the game...

Oddly enough, if I remove file logging OR switch to Clang, all those hits are gone. MS's Clang requires some slight changes to the code though, so there it is. v140 still compiles fine, of couse.

### DIFF
--- a/source/d3d8types.hpp
+++ b/source/d3d8types.hpp
@@ -5,6 +5,7 @@
 
 #pragma once
 
+struct IUnknown;
 #include <d3d9.h>
 
 #define D3DFMT_W11V11U10 65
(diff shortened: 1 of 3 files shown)


source/d3d8types.hpp
```diff
@@ -5,6 +5,7 @@
 
 #pragma once
 
+struct IUnknown;
```
crosire: Would prefer a `#include <unknwn.h>` after `#include <d3d9.h>` instead. Should work the same.

CookiePLMonster: It doesn't - with `#include <unknwn.h>` either above or below `d3d9.h` the error is same.
Apparently it's a pretty well known issue with Clang compilation so far and people are working this around that way with success. I myself have used this workaround in the past without any side effects.

crosire: Unfortunate.

PR #11, opened 2017-02-16
optional log

ThirteenAG:
Just a way to disable log output.

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -66,7 +66,9 @@ ULONG STDMETHODCALLTYPE Direct3DDevice8::Release()
 
 	if (myRef == 0 && ref != 0)
 	{
+#ifndef D3D8TO9NOLOG
 		LOG << "Reference count for 'IDirect3DDevice8' object " << this << " (" << ref << ") is inconsistent." << std::endl;
+#endif
 	}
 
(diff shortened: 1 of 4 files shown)


source/d3d8to9_device.cpp
```diff
@@ -561,8 +573,9 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::CopyRects(Direct3DSurface8 *pSourceSu
 
 		if (FAILED(hr))
 		{
+#ifndef D3D8TO9NOLOG
 			LOG << "Failed to translate 'IDirect3DDevice8::CopyRects' call from '[" << desc_source.Width << "x" << desc_source.Height << ", " << desc_source.Format << ", " << desc_source.MultiSampleType << ", " << desc_source.Usage << ", " << desc_source.Pool << "]' to '[" << desc_destination.Width << "x" << desc_destination.Height << ", " << desc_destination.Format << ", " << desc_destination.MultiSampleType << ", " << desc_destination.Usage << ", " << desc_destination.Pool << "]'!" << std::endl;
-			break;
+#endif			break;
```
crosire: You're missing a line feed here =)

ThirteenAG: Oh yeah, didn't notice that :)

PR #13, opened 2017-02-17
Update to fix graphic issues with Haegemonia and enhance compatibility

elishacloud:
- Removed "device->SetFVF(D3DFVF_XYZ);" line from Direct3D8::CreateDevice.  This fixes an issue with game Haegemonia Legions of Iron.  It seems there is no need to declare a default vertex here.   According to Microsoft documentation D3D8 applications should  call SetVertexShader API before using a vertex and SetFVF  already called when an application calls SetVertexShader.  https://msdn.microsoft.com/en-us/library/windows/desktop/bb204851(v=vs.85).aspx#Vertex_Declaration_Changes
 - Minor updates to convert_caps and convert_surface_desc to simplify code.
 - Updates to convert_present_parameters for better d3d9 copmatibility.
 - Updated logic for when MultiSampleType should be set to D3DMULTISAMPLE_NONE.
 - Changed PresentationInterval to default to D3DPRESENT_INTERVAL_ONE, the normal d3d9 default and makes more copmatible with d3d9 when D3DSWAPEFFECT_COPY_VSYNC was used.  According to Microsoft D3DPRESENT_INTERVAL_IMMEDIATE did nothing in fullscreen on d3d8 when D3DSWAPEFFECT_COPY_VSYNC was used.  So this is attempting to replicate that behavor.
 - Minor updates to the code slightly for when SwapEffect is set to D3DSWAPEFFECT_COPY.
 - Added some documentation to the code

### DIFF
--- a/source/d3d8types.cpp
+++ b/source/d3d8types.cpp
@@ -69,11 +69,11 @@ void convert_caps(D3DCAPS9 &input, D3DCAPS8 &output)
 {
 	CopyMemory(&output, &input, sizeof(output));
 
-	output.Caps2 |= D3DCAPS2_CANRENDERWINDOWED;								// Tell application that windows mode is supported
+	output.Caps2 |= D3DCAPS2_CANRENDERWINDOWED;								// Tell application that window mode is supported
 	output.RasterCaps |= D3DPRASTERCAPS_ZBIAS;								// Tell application that z-bias is supported
 	output.StencilCaps &= ~D3DSTENCILCAPS_TWOSIDED;							// Remove unsupported StencilCaps type
-	output.PixelShaderVersion = D3DPS_VERSION(1, 4);						// Set defaul PixelShaderVersion to 1.4 for D3D8 compatibility
-	output.VertexShaderVersion = D3DVS_VERSION(1, 1);						// Set defaul VertexShaderVersion to 1.1 for D3D8 compatibility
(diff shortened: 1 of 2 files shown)


source/d3d8types.cpp
```diff
@@ -98,40 +94,65 @@ void convert_surface_desc(D3DSURFACE_DESC &input, D3DSURFACE_DESC8 &output)
...
-	if (output.SwapEffect == D3DSWAPEFFECT_COPY_VSYNC)
+	// MultiSampleType must be D3DMULTISAMPLE_NONE unless SwapEffect has been set to D3DSWAPEFFECT_DISCARD.
+	// https://msdn.microsoft.com/en-us/library/windows/desktop/bb172588(v=vs.85).aspx
+	// Check for D3DMULTISAMPLE_NONMASKABLE and change it to D3DMULTISAMPLE_NONE for best D3D8 compatibility.
+	if (input.SwapEffect != D3DSWAPEFFECT_DISCARD || input.MultiSampleType == D3DMULTISAMPLE_NONMASKABLE)
 	{
-		output.SwapEffect = D3DSWAPEFFECT_COPY;
+		output.MultiSampleType = D3DMULTISAMPLE_NONE;
+	}
+	else
```
crosire: I rather prefer avoiding the else statements here by moving that code in front of the if as it was before. Other than that it looks great.

elishacloud:
Updated to remove else statements and move variable assignment above.

elishacloud:
Fix varable name in convert_present_parameters

crosire:
Perfect!

PR #16, opened 2017-04-29
Make d3dx9_xx.dll load dynamically

elishacloud:
d3dx8_43.dll was loaded statically before.  Some systems have different versions of this dll.  Changed it to load dynamically based on the version available on the system.

Note: This is also done by other programs.  Check out the log from ts3overlay [here](https://pastebin.com/s8RWZExL).

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -8,6 +8,99 @@
 #include <assert.h>
 #include <d3dx9shader.h>
 
+// D3DXLoadSurfaceFromSurface declaration for d3dx9_xx.dll
+typedef HRESULT(__stdcall *D3DXLoadSurfaceFromSurfaceType)(
+	_In_       LPDIRECT3DSURFACE9 pDestSurface,
+	_In_ const PALETTEENTRY       *pDestPalette,
+	_In_ const RECT               *pDestRect,
+	_In_       LPDIRECT3DSURFACE9 pSrcSurface,
+	_In_ const PALETTEENTRY       *pSrcPalette,
+	_In_ const RECT               *pSrcRect,
+	_In_       DWORD              Filter,
+	_In_       D3DCOLOR           ColorKey
+	);
+D3DXLoadSurfaceFromSurfaceType D3DXLoadSurfaceFromSurfacePtr = nullptr;
+
+// D3DXAssembleShader declaration for d3dx9_xx.dll
+typedef HRESULT(__stdcall *D3DXAssembleShaderType)(
+	_In_        LPCSTR        pSrcData,
+	_In_        UINT          SrcDataLen,
+	_In_  const D3DXMACRO     *pDefines,
+	_In_        LPD3DXINCLUDE pInclude,
+	_In_        DWORD         Flags,
+	_Out_       LPD3DXBUFFER  *ppShader,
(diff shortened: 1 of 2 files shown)


crosire:
I don't think this is necessary. D3DX is deprecated and won't be updated anymore (https://blogs.msdn.microsoft.com/chuckw/2013/08/20/living-without-d3dx/), meaning `d3dx9_43.dll` is the last version of that DLL that will ever exist. It's also the version that is installed by the last DirectSetup. From Windows 8 onward DirectX is part of the operating system and does not include D3DX, so the only way to get that DLL is to install via the DirectSetup from 2010 which contains this specific version of the DLL.
Unfortunately there don't seem to be any alternatives to assemble D3D8 shaders from assembler code, other than the deprecated D3DX, which is why it's in use here.

elishacloud:
Yes, you are probably right.  However I use this project inside a larger project and I did not want this dll to be static because the dll is not always needed for me.  I suspect others may do the same thing.

Also on my test system (Windows 7) I only have `d3dx9_32.dll` and other users who use my project my not have the latest dll also.  I don't want them to be required to update their DirectX libraries to use my project.

If you want I can change the number to start counting down at a smaller number rather than 99, like we can start at 50.  I did see some references to `d3dx9_47.dll` on the web, but they could be bogus.

crosire:
As soon as you install a DirectX game via Steam it will almost always install the DirectX libraries including `d3dx9_43.dll` too. The only systems I saw without that file were usually freshly set up.
I don't feel like merging this, but that doesn't have to stop you from using this change in your own fork.

elishacloud:
Ok, understood.  Thanks for looking at it.  I will just have this on my branch.

CookiePLMonster:
While I don't like the way this is trying for different DLL versions, loading it dynamically could still be beneficial since not everyone has DX redists installed (you'd be surprised how many people do **not** have them!) and I don't think shaders were a common thing in d3d8 days - so most games would be unaffected.

elishacloud:
Ok, let me change that to dynamically load `d3dx9_43.dll`.

source/d3d8to9_device.cpp
```diff
@@ -52,29 +52,15 @@ bool LoadD3dx9()
...
-			strcpy_s(d3dx9name, "d3dx9_");
-			char buffer[11];
-			_itoa_s(x, buffer, 10);
-			strcat_s(d3dx9name, buffer);
-			strcat_s(d3dx9name, ".dll");
-
-			// Load dll
-			dllHandle = LoadLibrary(d3dx9name);
-		}
+		dllHandle = LoadLibrary("d3dx9_43.dll");
```
CookiePLMonster: Unicode please? =) This is the first ever line which will be affected by multibyte/Unicode switch so while I'm not a fan of using MS macroes for multibyte/Unicode, ideally you could replace those two lines with

```c++
HMODULE dllHandle = LoadLibrary(TEXT("d3dx9_43.dll"));
```

and support both character encodings at once. Still, project should probably end up being switched to Unicode as it's the way Windows-only projects should be handled these days to be honest.

elishacloud:
Ok, good call.  Just updated this to support Unicode, as requested.

crosire:
Alright. You got me convinced. I'll clean this up a little in a follow-up commit though.

PR #20, opened 2017-05-06
Alternate refcounting method

### LINKED ISSUE #19: Crashing when exiting the game Rayman 3: Hoodlum Havoc
Unfortunately one of the check-ins today is causing a crash.  It seems to be this check-in [a7b2a63](https://github.com/crosire/d3d8to9/commit/a7b2a632ca63f68b7f5ac7c46d52006618516e2d) "Remove standalone DirectX SDK dependency".  If I revert this check-in the crash goes away.

The updated code for the dynamic loading of d3d9_43.dll in this check in seems fine and the three API calls `D3DXAssembleShader`, `D3DXDisassembleShader` and `D3DXLoadSurfaceFromSurface` seem to work ok.  I am guessing it is something in (or missing in) "d3dx9.hpp".  Somehow removing the SDK "d3dx9shader.h" and replacing it with "d3dx9.hpp" is causing an issue.

The crash only happens when the game is exiting.  It seems to run fine (normal) otherwise.

I did not have time to fully troubleshoot it.  Here is the crash I am seeing:
crash

CookiePLMonster:
With this change, underlying d3d9 resources don't have their refcounts changed by wrapper's `AddRef` and `Release` and instead they only get released once at wrapper destruction time. Instead of returning underlying resource's refcount from wrapper functions, now wrapper's refcounts are returned. This fixes cases where the game is doing something like

```c++
while(resource->Release() > 0);
```

Might also be a micro optimization since it issues less d3d calls (albeit refcount ones should be nearly no-ops).

Fixes #19

### DIFF
--- a/d3d8to9.vcxproj
+++ b/d3d8to9.vcxproj
@@ -18,14 +18,14 @@
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
     <PlatformToolset>v140</PlatformToolset>
-    <CharacterSet>MultiByte</CharacterSet>
+    <CharacterSet>Unicode</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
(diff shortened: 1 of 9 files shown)


CookiePLMonster:
Also switched the project to Unicode because... why not? W versions of WinAPI functions should be preferred these days and the only affected line is

```c++
const HMODULE module = LoadLibrary(TEXT("d3dx9_43.dll"));
```

PR #23, opened 2017-05-19
Message box when failed to load d3dx9_43.dll

HerMajestyDrMona:
Added a message box when d3d8to9 is unable to load d3dx9_43.dll. It asks users to download the Microsoft DirectX® End-User Runtime, and also allows to ignore the warning when the "cancel" button is clicked.


crosire:
The messagebox text shows a different link compared to what's actually opened. I'd not show the link on the messagebox at all.

Also, don't hardcode the language, better use a language-neutral link like the following:
https://www.microsoft.com/download/details.aspx?id=35

HerMajestyDrMona:
Thanks for reviewing it :) I don't know why I used the offline-installer link in the message box, but not in the ShellExecute function. Now ShellExecute will open the online installer link.

Issue #25, opened 2017-06-08
Indiana Jones And The Emporer's Tomb missing textures


Radio30:
Hi, I've just been using d3d8to9 for the first time and am getting missing textures through out the game:

Before: http://imgur.com/zujLRLe
After: http://imgur.com/ZO0PU7B

Also the in-game movies are just a black screen with audio, the game never crashes and works fine otherwise. 
Is there a log file somewhere i'm missing that might tell me whats causing the missing textures?
cheers

elishacloud:
I purchased the game on [GOG](https://www.gog.com/game/indiana_jones_and_the_emperors_tomb) and was able to see the same issue of the missing textures.  When I try other converters, like the one [here](http://enbdev.com/convertor_dx8_dx9_v0035.htm), it works fine.  I believe this is the same issue I reported on [Hitman 2 Silent Assassin](https://github.com/crosire/d3d8to9/issues/17).

BTW: this game is on sale for $1.49 on [GOG](https://www.gog.com/game/indiana_jones_and_the_emperors_tomb) until June 27, if someone wants to try and reproduce this issue.

elishacloud:
Ok, I figured out why the textures are missing. It seems that _Indiana Jones and the Emperor's Tomb_ (indy.exe) is calling `CreateTexture()` 468 times when the game first starts.  After that it tries calling `SetTexture()`.  From what I can tell it is looping and calling `SetTexture()` around 192 times or so each loop.

However the issue is that  around 42 of the times each loop `SetTexture()` is called with a `nullptr`.  So somehow `Direct3DTexture8` is getting deleted between the time it is created and the time indy.exe calls the `SetTexture()` API.

A few things to note about the `SetTexture()` API calls:
1. `Stage` = 0, 1, 2, 3 are always called in order.
2. `Stage` = 1, 2 and 3 seem to fail for the first 10 to 14 times each loop, then they start working for the rest of the loop.
3. `Stage` = 0 always works.
4. `Stage` = 1 is always a D3DRTYPE_CUBETEXTURE, whereas all the other stages are always D3DRTYPE_TEXTURE.

I tested this with [Hitman 2 Silent Assassin](https://github.com/crosire/d3d8to9/issues/17) and got the same results where `SetTexture()` is called with a `nullptr` so it seems to be the same issue.

I added a few extra log entries and attached them if someone wants to look at it.

Here is an example:
```text
.....
[...]

Issue #17, opened 2017-04-29
Missing textures in Hitman 2 Silent Assassin


elishacloud:
There seems to be missing textures on [Hitman 2 Silent Assassin](https://www.gog.com/game/hitman_2_silent_assassin) when using this wrapper/converter.

Here is what I get with [d3d8to9](https://github.com/crosire/d3d8to9):
![hitman2_d3d8tod9](https://cloud.githubusercontent.com/assets/25625456/25559308/11d56066-2cec-11e7-8c71-2675e21c87a5.png)

Here is what it is supposed to look like:
![hitman2](https://cloud.githubusercontent.com/assets/25625456/25559312/22e401b4-2cec-11e7-9931-51a6b7cba461.png)

I think this problem is different than the problem with [Rayman 3: Hoodlum Havoc](https://github.com/crosire/d3d8to9/issues/14) because when I try other converters, like the one [here](http://enbdev.com/convertor_dx8_dx9_v0035.htm), it works fine.  I am hoping someone can look into this as I need this converter to play the game in Windows 10.

CookiePLMonster:
Have you tried comparing pix frames from this wrapper and the one from Boris/ENBSeries? While those are useless for anyone but you (frames can't be viewed from a PC with a different driver/GPU model), maybe you'll spot something obvious in the render states (or texture sampler settings) for 47's draw call.

elishacloud:
Thanks for the suggestion!  However I an new to DirectX programming (and C/C++ programming for that matter).  I am not sure how to compare pix frames.  I don't have the source code for the Boris/ENBSeries converter.

CookiePLMonster:
You'd need to use it to capture frames from ingame and then most likely compare render states during the draw call - no need for the source.

elishacloud:
Ok, I started looking at this in PIX for Windows using this guide [here](https://blogs.msdn.microsoft.com/manders/2006/12/15/a-painless-introduction-to-pix-for-windows/).  So far I have found three differences between d3d8to9 and the Boris/ENBSeries converter.

1. In the `SetVertexShader` function it seems that the Boris/ENBSeries converter calls `SetFVF` and `SetVertexDeclaration` before it calls `SetVertexShader`.  Whereas d3d8to9 calls them in reverse order.

2. In the `SetRenderTarget` function the Boris/ENBSeries converter does not call `Release` and only calls `AddRef` once.  Whereas d3d8to9 calls `Release` and calls `AddRef` more than once.

[...]

PR #27, opened 2017-06-25
Update CreateVertexShader to fix missing textures

elishacloud:
- Updated `CreateVertexShader` to fix missing textures in some games such as [Hitman 2 Silent Assassin](https://github.com/crosire/d3d8to9/issues/17) and [Indiana Jones and the Emperor's Tomb](https://github.com/crosire/d3d8to9/issues/25).
- Updated return value for `CopyRects`, `CreateVertexShader`, `GetVertexShaderDeclaration` and `CreatePixelShader`.  Removed `E_FAIL` and `E_NOTIMPL` and replaced them with `D3DERR_INVALIDCALL`.  These Direct3D APIs don't list `E_FAIL` and `E_NOTIMPL` as supported return values in the Microsoft documentation.
- Fixed `SetRenderState` to call the correct `D3DRENDERSTATETYPE` when `D3DRS_ZBIAS` is used.


PatrickvL:
Nice work! As for the zbias to depthbias mapping - I choose the same mapping a few years ago - see https://github.com/PatrickvL/Dxbx/blob/master/Source/Delphi/src/DxbxKrnl/EmuD3D8/uConvert.pas#L436 this might contain other mappings that could be useful for this project (although to be honest, these weren't tested thoroughly, so I'm not 100% sure those mappings in there are entirely correct, but at least it's a reference that can be looked at, right?)

elishacloud:
Yes, changing `D3DRS_ZBIAS` to `D3DRS_DEPTHBIAS` may not be a perfect change.  However if you look at `Direct3DDevice8::GetRenderState` it appears that this is what d3d8to9 is expecting.  There are other ways of handling this also.  Take a look at the comments [here](https://gamedev.stackexchange.com/questions/28244/unexpected-results-with-d3d9-depth-bias) and [here](https://software.intel.com/en-us/articles/alternatives-to-using-z-bias-to-fix-z-fighting-issues/).

It seems that the [Boris/ENBSeries](http://enbdev.com/convertor_dx8_dx9_v0035.htm) converter does something similar to what d3d8to9 does.  Here is what ENBSeries does:
```c++
case D3DRS_ZBIAS:
	Biased = static_cast<FLOAT>(Value) * -0.0f;
	Value = *reinterpret_cast<const DWORD *>(&Biased);
	State = D3DRS_DEPTHBIAS;
	SetRenderState(State, Value);
	break;
```
Whereas d3d8to9 does this:
```c++
case D3DRS_ZBIAS:
	Biased = static_cast<FLOAT>(Value) * -0.000005f;
	Value = *reinterpret_cast<const DWORD *>(&Biased);
	State = D3DRS_DEPTHBIAS;
	SetRenderState(State, Value);
	break;
```
I am not quite sure which one is better.  All five games I tested showed the same results no matter which one of these methods I used.  So I took the simplest path.

As far as the fix for `CreateVertexShader`, I did notice that d3d8to9 will initialize all three registers: 'oD', 'oT' and 'r'.  However the ENBSeries  only initializes the 'oT' registers.  I did not see a need to initialize the 'oD' and 'r' registers but I left them there as it seemed to work.  Though I wonder if initializing these other two registers are really needed.

crosire:
There was a game (the first Max Payne I think) that would only write to certain components of some registers, leaving the rest uninitialized. Passing that code to the D3DX assembler would fail with an error, which is why the initialization happens. I don't know if this works even if `c0` is not declared, since it's a constant either way. But it probably will, so I guess it's OK.

elishacloud:
I tested both Max Payne and Max Payne 2.

Max Payne fails both with and without this change and never calls `CreateVertexShader`.  Something weird is happening with Max Payne because it appears that all APIs are already hooked and it won't launch if I try to attach a debugger to it or run it with PIX.  I cannot get Max Payne to work with d3d8to9 even with older versions of d3d8to9.

Max Payne 2 seems to run fine with d3d8to9 with this change.

Let me know if there is some other issue with this change.  It seems to work fine on all the games I have tested.

crosire:
Alright. Looks good then.

Issue #28, opened 2017-06-28
Vertex shader translation does not work as expected (Star Wars Republic Commando)


metallicafan212:
So I've been following this project silently for about a year and a half for now as it interested me in the project I'm working on.

I have no c++ knowledge of any sorts, since I am primarily a Java programmer, but I hope I can help with one issue with vertex shaders.

For the longest time, the hardware shaders (Unreal Engine 2 combines vertex and pixel shaders) for some of the effects have not been being rebuilt properly. In d3d8, the shaders work fine, but in d3d8to9 it rebuilds the vertex shaders with wrong instructions.

Take for example: the hardware shader DynamicHologram, this is how it appears in editor
(I'm using quote since the code tag does not want to work with it)

> vs.1.1
> 
> ; Get the Position
> mov r3, v0
> 
> ; Add big rare flicker
> mul r3.z, r3, c21
> 
> ; Add small frequent flicker
> add r3.xyz, r3, c22
> 
> ////////// Distort /////////
> 
> // Position distort
> mul r1, v0, c16.x
> add r1.xy, r1.zx, r1.yz
> 
> // Time distort
> mov r2, c17
> mad r1, r2.x, c16.y, r1
> frc r1.xy, r1        // 0 to 1
> 
> // Turn linear into sawtooth
> add r1, r1, -c20.y    // -.5 to .5
> max r7, r1, -r1      // .5 to 0 to .5
> 
> // Distort
> mul r4, r7, c16.zzww // z = Amp  w = 0
> mul r4, r4, r3
> mov r4.z, r3.z
> mov r4.w, c20.z
> 
> ////////// Scan /////////
> mov r5, r3
> mov r5.z, c18.y
> 
> ////////// Combine /////////
> 
> // If z > Scan then Distort
> sge r0.x, r3.z, c18.x
> 
> // if z < distort and > Scan then scan
> slt r0.y, r3.z, c18.x
> sge r0.z, r3.z, c18.y
> mul r0.y, r0.y, r0.z
> 
> // if z < Scan then scan
> slt r0.z, r3.z, c18.y
> 
> mul r1, r4, r0.x
> mad r1, r5, r0.y, r1
> mad r3, r3, r0.z, r1
> 
> ////////// Transform /////////
> 
> ;transform position
> dp4 r4.x, r3, c0
> dp4 r4.y, r3, c1
> dp4 r4.z, r3, c2
> dp4 r4.w, r3, c3
> 
[...]

PR #30, opened 2017-07-02
Fix shader translation for Star Wars Republic Commando

elishacloud:
Added `(?![0-9])` to ensure that all the numbers get picked up in a multi-digit register name when translating vector and pixel shaders.

This fixes the issue reported in [#28](https://github.com/crosire/d3d8to9/issues/28).

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -1386,9 +1386,9 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::CreateVertexShader(const DWORD *pDecl
 
 		SourceCode = std::regex_replace(SourceCode, std::regex("    \\/\\/ vs\\.1\\.1\\n((?! ).+\\n)+"), "");
 		SourceCode = std::regex_replace(SourceCode, std::regex("(oFog|oPts)\\.x"), "$1 /* removed swizzle */");
-		SourceCode = std::regex_replace(SourceCode, std::regex("(add|sub|mul|min|max) (oFog|oPts), ([cr][0-9]+), (.+)\\n"), "$1 $2, $3.x /* added swizzle */, $4\n");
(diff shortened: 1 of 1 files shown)


elishacloud:
Made some changes to the pull request.  This fixes the first 4 issues listed on this comment [here](https://github.com/crosire/d3d8to9/issues/31#issuecomment-312699542).

I also noticed that only one line in `CreateVertexShader` needed to be updated because these two [here](https://github.com/crosire/d3d8to9/pull/30/files#diff-366bc9823b5fedd9a67231f1896c4f8fR1389) and [here](https://github.com/crosire/d3d8to9/pull/30/files#diff-366bc9823b5fedd9a67231f1896c4f8fR1390) don't have any issues.  So I reverted the changes on these lines

This basically fixes six things:
1. An issue with vertex shader with Star Wars Republic Commando listed [here](https://github.com/crosire/d3d8to9/issues/28).  Fixed in [this](https://github.com/crosire/d3d8to9/pull/30/files#diff-366bc9823b5fedd9a67231f1896c4f8fR1391) line.
2. Missing (not removing of) modifiers. This is fixed in this line [here](https://github.com/crosire/d3d8to9/pull/30/files#diff-366bc9823b5fedd9a67231f1896c4f8fR1733).
3. Missed modifiers when swizzles are used. This is also fixed in this line [here](https://github.com/crosire/d3d8to9/pull/30/files#diff-366bc9823b5fedd9a67231f1896c4f8fR1733).
4. Dangling Swizzles when removing modifiers from constants.  This is fixed in this line [here](https://github.com/crosire/d3d8to9/pull/30/files#diff-366bc9823b5fedd9a67231f1896c4f8fR1735).
5. Possible dangling modifiers when removing modifiers from constants. This is fixed by moving [this](https://github.com/crosire/d3d8to9/pull/30/files#diff-366bc9823b5fedd9a67231f1896c4f8fR1733) line up in front of the other regex lines.
6. Fixing of the simple modifier `-` when `add` is used to prevent certain graphic glitches in games like _Star Wars Republic Commando_ and _Silent hill 2_.  Added [this](https://github.com/crosire/d3d8to9/pull/30/files#diff-366bc9823b5fedd9a67231f1896c4f8fR1734) line for that.

crosire:
Great work!

Issue #21, opened 2017-05-06
Back buffer reference counting


CookiePLMonster:
When investigating on why both d3d8to9 and ENB's d3d8 wrapper flat out crash with a dx8 thrash driver in Need for Speed III (from [THIS PATCH](http://veg.by/en/projects/nfs3/)) I noticed a couple issues, but the most important fact is that this code loves to do stuff like:

```c++
device->GetBackBuffer(0, &surface);
surface->Release();
global_myBackbuffer = surface;
```

Since d3d8to9 creates a new wrapper class in `GetBackBuffer`, the call to `Release` destroys it and any attempts to use this surface fail miserably.

Should this be considered severe API misuse and not handled, or is this something to fix on d3d8to9 side? I mean, this is obviously **not** how you should be managing reference counters but this doesn't seem to crash with a real dx8 device at all. Maybe device wrapper should account for such odd behaviour, store its backbuffer surface wrappers in an array and return those on demand? This should probably match real device behaviour more closely.

EDIT:
Said game also seems to crash consistently inside `Direct3DDevice8::SetCurrentTexturePalette` but that seems like a different issue. What's worse, it instantiates itself with `D3DCREATE_MULTITHREADED` and who knows how thread unsafe d3d8to9 device wrapper really is.

crosire:
While it's a bad idea it's not technically wrong to use the API like that, since the device has a guaranteed backbuffer instance for its entire lifetime. So yes, I guess we should cache the backbuffer in the device wrapper. I don't think we need to handle more than the first backbuffer though. It's rare more than one exists.

As for thread-safety: Most calls are implicitly handled by the critical section the D3D9 runtime acquires, since the same flag is passed on to it as well. However, some calls that modify states in d3d8to9, like `Direct3DDevice8::SetRenderTarget` are certainly not thread-safe.

CookiePLMonster:
I don't know what the D3D8 limit is, but D3D9 limit is 3 so it shouldn't be much more effort than handling only the first backbuffer.

Then there are swap chains, though... maybe d3d8 device should store handles to its implicit swap chains and use them for `D3D8Device::GetBackBuffer` and the others?

elishacloud:
[...]

PR #34, opened 2017-07-08
Add caching for Direct3D8 classes

elishacloud:
Added caching for the following classes:
- Direct3DSurface
- Direct3DTexture
- Direct3DCubeTexture
- Direct3DVolumeTexture
- Direct3DVolume
- Direct3DVertexBuffer
- Direct3DIndexBuffer
- Direct3DSwapChain8

Simplified reference counting by removing `RefCount`, `CurrentRenderTarget` and  `CurrentDepthStencilSurface`

This fixes the problems reported in issue [#21](https://github.com/crosire/d3d8to9/issues/21).


elishacloud:
Let me explain this change in more detail also since this is a rather large change.
## New 'Direct3DCache' class added
I created a new class called `Direct3dCache` to handle all the caching of the addresses.  Along with this I added two new files: direct3dcache.cpp and direct3dcache.h

This caching class has three functions that are exposed for each Direct3D8 class.  The function names are `SetDirect3D`, `GetDirect3D` and `DeleteDirect3D`.

Here is an example of the `Direct3DSurface8` class functions:
```c++
void SetDirect3D(Direct3DSurface8*, IDirect3DSurface9*);
Direct3DSurface8* GetDirect3D(IDirect3DSurface9*);
void DeleteDirect3D(Direct3DSurface8*);
```

The `Direct3dCache` class creates an array of vectors, one vector for each Direct3D8 class.  The index of each Direct3D8 vector is stored in a `static const` so that it can be referred to by name only:
```c++
static constexpr DWORD SurfaceVector = 0;
static constexpr DWORD TextureVector = 1;
static constexpr DWORD CubeTextureVector = 2;
static constexpr DWORD VolumeTextureVector = 3;
static constexpr DWORD VolumeVector = 4;
static constexpr DWORD VertexBufferVector = 5;
static constexpr DWORD IndexBufferVector = 6;
static constexpr DWORD SwapChainVector = 7;
```

PatrickvL:
Well done! However, these function names are a bit confusing. Better use a term like "HostResource" or something along those lines

elishacloud:
Good point!  The names were confusing.  I changed the names as follows:

- `Direct3DCache` changed to `AddressLookupTable`
- `MyDirect3DCache` changed to `ProxyAddressLookupTable`
- `SetDirect3D` changed to `SaveAddress`
- `GetDirect3D` changed to `FindAddress`
- `DeleteDirect3D` changed to `DeleteAddress`

I also changed the name of the files to `lookup_table.cpp` and `lookup_table.h` and moved these to the 'helpers' folder in the Visual Studio project.

My comments above have been modified to use the new names as well.  These changes should make the code easier to read.

Note: I did not want to use the word 'Resource' since that has a very specific meaning in a Visual Studio project.

PatrickvL:
Magnificent!

About the bugfix - since `DeleteDirect3D` previously called `GetDirect3D`, the test-results need to be revisited too, right?

elishacloud:
You are right I did not go back and retest all the games again after this last change.  I just did a quick smoke test on _Need for Speed III_.  However I am not too worried about the issue I fixed with `DeleteDirect3D` causing any issues since that function is not actually used.  It is only called if the class is deleted separately from the `DeleteMe` function, which currently does not happen.  If needed I can go back and retest all the games again.  :)

elishacloud:
I fixed an issue with `VolumeTexture` and `CubeTexture` I did not notice before and did a quick test of the 10 games again to ensure everything is working ok.

crosire:
I started a review and made some comments, but am not sure if you saw them (above).

elishacloud:
I am not seeing any of your comments for some reason.  Where should I be looking?

crosire:
Whoops. Forgot to click that big "Submit" button.

elishacloud:
Great ideas!  I made the requested changes.

source/d3d8to9_device.cpp
```diff
@@ -1060,6 +978,7 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::GetPaletteEntries(UINT PaletteNumber,
 }
 HRESULT STDMETHODCALLTYPE Direct3DDevice8::SetCurrentTexturePalette(UINT PaletteNumber)
 {
+	PaletteNumber = (PaletteNumber & 0xFFF) | (PC_NOCOLLAPSE << 12);
```
crosire: I'm curious as to why/how this works?

elishacloud: I am not quite sure.  I did some research and the `PALETTEENTRY` structure is documented [here](https://msdn.microsoft.com/en-us/library/windows/desktop/bb147253(v=vs.85).aspx).  It seems to crash anytime `SetCurrentTexturePalette` is called with any flag set other than `PC_NOCOLLAPSE`.  However it is not clear why it crashes.  I did find the comment below in the documentation, but that is all.  These flags do not seem well documented:

> The alpha intensity value for the palette entry. Note that as of DirectX 8, this member is treated differently than documented for Windows.

crosire: Yes, but those flags are for the `PALETTEENTY` structure, so I'm confused as to why adding it to the palette index changes things.

crosire:
FYI, I modified the code slightly so it uses templates instead of a bunch of function overloads to reduce code size/redundancy a bit.

source/d3d8to9_device.cpp
elishacloud: Yes, I see your point.  That is suppose to be the palette index not a `PALETTEENTRY` structure.  I did a bit more research here and found some things that may shed light on this issue.

First of all it seems that 64bit Windows does not support `PALETTEENTRY`.  I found [this](https://msdn.microsoft.com/ja-jp/subscriptions/dd144877.aspx) documentation on a Microsoft site:

> Number of entries in the system palette. This index is valid only if the device driver sets the RC_PALETTE bit in the RASTERCAPS index and is available only if the driver is compatible with 16-bit Windows.

Notice the last line "is available only if the driver is compatible with 16-bit Windows".  As you know 64bit Windows is **not** compatible with 16-bit Windows.  I think this is trying to say that palettes are only supported if 16-bit Windows is supported.  I can verify this as all attempts to use the `PALETTEENTRY` on my 64bit system (even with GDI) has resulted in a crash.

Furthermore after looking closely at the `PC_NOCOLLAPSE` documentation I found [here](https://msdn.microsoft.com/en-us/library/cc250401.aspx) it says:

> Specifies that the color be placed in an unused entry in the system palette instead of being matched to an existing color in the system palette. If there are no unused entries in the system palette, the color is matched normally. Once this color is in the system palette, colors in other logical palettes can be matched to this color.

Notice it says "If there are no unused entries in the system palette, the color is matched normally."  It seems that the `PC_NOCOLLAPSE` will match the color normally if no system entry palettes exist, which is the case on my system.

I also read somewhere (though I cannot find it now) that the `PaletteNumber` used by `SetPaletteEntries` somehow corresponds to the `PALETTEENTRY`.

So as far as I can tell the reason `PC_NOCOLLAPSE` is working here is because Direct3D9 sees that there is no system palettes available and just matches this color normally.

Issue #14, opened 2017-02-19
Rayman 3: Hoodlum Havoc - Deformed Models and Incorrect Textures when Hardware Vertex Processing is enabled


RibShark:
I'm trying to play Rayman 3: Hoodlum Havoc with d3d8to9, and I've run into a bit of an issue. When in any area of the game apart from the menu, various parts of the scenery can be highly deformed or appear completely invisible:
![image](https://cloud.githubusercontent.com/assets/1957489/23103518/27a56396-f6b4-11e6-96bc-0157f69e7f22.png)

For comparison, this scene is supposed to look like this:
![image](https://cloud.githubusercontent.com/assets/1957489/23103606/596d5536-f6b5-11e6-8674-72b833667cd9.png)

On some in game objects, the textures used are also incorrect (not sure if it's using the wrong texture or it is mapped incorrectly onto the model). Additionally, one in game object can even appear to be using the model of another object (the gem in this picture is actually a crown):
![image](https://cloud.githubusercontent.com/assets/1957489/23103583/e16c173e-f6b4-11e6-8606-779174e0d927.png)

For comparison, this scene is supposed to look like this:
![image](https://cloud.githubusercontent.com/assets/1957489/23103617/8c5028b6-f6b5-11e6-9ca3-e2449864de95.png)

This can be fixed while still using d3d8to9 by opening the file C:\Windows\Ubisoft\Ubi.ini and changing the line TnL=1 to TnL=0. After analysing the EXE, it seems that all this is doing is setting d3d to use Software Vertex Processing (but I could be wrong). The downside of this is that the framerate tanks to 20fps or less.

I'm using a GTX 970 with the latest drivers, if that's any help.

elishacloud:
Can you try with the following versions to see if they make any difference?

[d3d8v14.zip](https://github.com/crosire/d3d8to9/files/786020/d3d8v14.zip)
[d3d8v15.zip](https://github.com/crosire/d3d8to9/files/786021/d3d8v15.zip)

RibShark:
Thanks for looking into this! Neither DLL acts any differently, unfortunately.

elishacloud:
I purchased the game on [GOG](https://www.gog.com/game/rayman_3_hoodlum_havoc) and was able to see that some of the textures are missing, though it did not look exactly like your screenshots so I am not completely sure it is the same issue.

[...]

PR #35, opened 2017-07-09
Fix Vertex and Index Buffer Locks that don't use D3DUSAGE_WRITEONLY

elishacloud:
Update the Lock() functions for Vertex and Index Buffers so that they ignore the `D3DLOCK_DISCARD` flag if the `D3DUSAGE_WRITEONLY` flag is not set.  Fixes the missing texture issue with [Raymond 3](https://github.com/crosire/d3d8to9/issues/14).


crosire:
Awesome!

PR #38, opened 2017-07-15
Check for Palette support

elishacloud:
I updated the palette code a little bit.  With this update the code will first check if palettes are supported.  The `PC_NOCOLLAPSE` flag will only be set  if palettes are **not** supported.  Also this update will do some extra error checking on the `SetPaletteEntries`, `GetPaletteEntries` and `GetCurrentTexturePalette`.

Note: as far as I can tell there is no way `GetCurrentTexturePalette` will ever work if palettes are not supported.  It always seems to crash.  This code should prevent crashes from any game that properly uses any of these four palette APIs (whether palettes are supported or not).


PatrickvL:
The palette check via raster caps doesn't work on all systems when checked under Direct3D 8.

elishacloud:
Interesting!  But, one thing to note here is that we are running under Direct3D 9, not Direct3D 8.  This may be an issue with Direct3D 8 but do we know if the same issue exists in Direct3D 9?   I know the OS does handle palettes differently between Direct3D 8 and 9.  For example in Direct3D 8 you can call some of the palette functions even when it is not supported without it crashing, whereas in Direct3D 9 if you try and run any function related to palettes it will crash if palettes are not supported.  It is possible that the same update done in [Cxbx](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded) would work on Direct3D 8 but cause a crash in Direct3D 9.

When I was testing this I did see that `RASTERCAPS` would return incorrect results if I put this check under the `Direct3D8` class constructor.  This is part of the reason I added this under the `Direct3DDevice8` class constructor because by this point we have already created an `IDirect3DDevice9` device so Windows will see us as a Direct3D 9 application.  In my tests this made it so that `RASTERCAPS` result was correct.

If you have any more data about when the `RASTERCAPS` call could be incorrect under Direct3D 8 I would be interested to test it with this update and see if we have the same issue.

elishacloud:
One more comment here is that [Cxbx](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded) called `RASTERCAPS` _before_ creating the `Direct3DDevice`.  I found that to be troublesome also.  However when calling this _after_ creating the `Direct3DDevice` it seemed to be reliable based on my tests.

PatrickvL:
Thanks for the feedback. I'm sorry I don't have more detail on this. I guess it would need testing on a wide range of cards and drivers.

Issue #39, opened 2017-08-05
[GTA III/VC] Crash error when changing resolution


gwet18:
III:
![1](https://user-images.githubusercontent.com/26855606/28999365-cf24ae08-7a4c-11e7-9997-ca06f15c4e1f.png) then ![2](https://user-images.githubusercontent.com/26855606/28999366-df11da98-7a4c-11e7-8266-d57a6e30f266.png)

VC:
![3](https://user-images.githubusercontent.com/26855606/28999369-e9123434-7a4c-11e7-88bd-d72c550cfc3c.png) then ![4](https://user-images.githubusercontent.com/26855606/28999371-f0623d24-7a4c-11e7-8aab-b63e6404205f.png)

LethalManBoob:
Same crashes here on serious sam. Expect its also when i alt tab as the game needs to recall directx. This crashes it.

elishacloud:
The issue with GTA VC is that after the resolution changes it calls `CreateDevice` to create another Direct3D device however when the [d3d8to9](https://github.com/crosire/d3d8to9/) wrapper tries to create the device Windows is returning error -2005530520 (88760868) which basically means the device is lost `D3DERR_DEVICELOST`.  The solution is to simply try creating the device again.  It seems to always work the second time.

I am not sure if the same fix will work on GTA III and Serious Sam since I don't own those games.  Try with [this](https://github.com/crosire/d3d8to9/files/1219596/d3d8.zip) file here and let me know how it works.

gwet18:
I think it doesn't work either, same errors..

elishacloud:
Yes, you are right.  There is more than one issue.  Apparently GTA does not always close the current device before trying to create a new one.  Try [this](https://github.com/crosire/d3d8to9/files/1219873/d3d8.zip) one out.

gwet18:
Toggling resolutions in III seems to be fine, in VC gives a different error now:
![default](https://user-images.githubusercontent.com/26855606/29246789-7b12e4a4-800c-11e7-8613-e4e6f7a2fd73.png)
[crashdump](https://github.com/crosire/d3d8to9/files/1220318/gta-vc.exe.3468.zip)

elishacloud:
I am running VC and it works fine for me.  Can you tell me what OS you are using, what version of VC you are using and what type of video card you have?

[...]

PR #40, opened 2017-08-15
Fix device lost error when creating device to fix issue with GTA III/VC

elishacloud:
This fix has several changes it in:

1. Error with lost device when CreateDevice is called:

The core issue with GTA III/VC mentioned in #39 is that GTA would try and create a new device when the old device was still there.  This caused it to return error `D3DERR_DEVICELOST`.  However GTA does not know how to properly handle this error and simply crashed.  The fix is to call `Release()` to release the previous device when a device lost error is seen so that a new device can be created.

It also seems that after the new device is created then GTA VC will call `Release()` and release the old device.  So this update will release the Direct3D9 device.  When GTA VC calls `Release()` then the Direct3D8 device (d3d8to9) will get released.

2. Invalid display type listed with 0-bits of color:

When testing GTA VC I noticed it showed a device with 0-bits of color (should only show 16-bit and 32-bit color).  I traced the issue to d3d8to9 using an unsupported `D3DFMT_A2R10G10B10` adapter format.  See [here](https://msdn.microsoft.com/en-us/library/ms886236.aspx) for documentation on Direct3D8 supported adapter formats.  Also you can look at 'd3d8types.h' from the DirectX SDK to see supported Direct3D8 formats.

Note: don't be confused by `D3DFMT_A2R10G10B10` and `D3DFMT_A2B10G10R10`, while similar these are different.  I also noticed that `D3DFMT_A2R10G10B10` did not exist in d3d8to9's 'd3d8types.cpp'.  I am not sure it should be added or not.

3. Issue with palette when palettes are not supported:

Also while I was testing this fix I found an issue in _Indiana Jones and the Emperor's Tomb_ where it would crash because of how pixels are handled in d3d8to9 (from this change [ e0d43b5](https://github.com/crosire/d3d8to9/commit/e0d43b56dfb86a62e1d39804fb0e86409fb62ce4) for the crash in _Need for Speed III_).  To fix this I changed the code to simply return `D3DERR_INVALIDCALL` when `SetCurrentTexturePalette` is called and palettes are not supported.  This fixes both _Indiana Jones and the Emperor's Tomb_ and _Need for Speed III_.

4. Logging will stop when Direct3DCreate8 is called more than once:

In some games `Direct3DCreate8()` is called more than once.  D3d8to9 tries to open the log file each time `Direct3DCreate8` is called.  This caused logging to stop at that point.  This update fixes that issue by checking if the log is already open before opening it again.

### DIFF
--- a/source/d3d8to9.cpp
+++ b/source/d3d8to9.cpp
@@ -18,7 +18,10 @@ std::ofstream LOG;
 extern "C" Direct3D8 *WINAPI Direct3DCreate8(UINT SDKVersion)
 {
 #ifndef D3D8TO9NOLOG
-	LOG.open("d3d8.log", std::ios::trunc);
+	if (!LOG.is_open())
+	{
+		LOG.open("d3d8.log", std::ios::trunc);
+	}
 
 	if (!LOG.is_open())
 	{
--- a/source/d3d8to9.cpp
+++ b/source/d3d8to9.cpp
@@ -18,13 +18,16 @@ std::ofstream LOG;
 extern "C" Direct3D8 *WINAPI Direct3DCreate8(UINT SDKVersion)
 {
 #ifndef D3D8TO9NOLOG
+	static bool LogMessageFlag = true;
+
 	if (!LOG.is_open())
 	{
 		LOG.open("d3d8.log", std::ios::trunc);
 	}
 
-	if (!LOG.is_open())
+	if (!LOG.is_open() && LogMessageFlag)
 	{
+		LogMessageFlag = false;
 		MessageBox(nullptr, TEXT("Failed to open debug log file \"d3d8.log\"!"), nullptr, MB_ICONWARNING);
 	}
 
@@ -65,7 +68,6 @@ extern "C" Direct3D8 *WINAPI Direct3DCreate8(UINT SDKVersion)
 				return nullptr;
 			}
 		}
-
 	}
 
 	return new Direct3D8(d3d);
--- a/source/d3d8to9_base.cpp
+++ b/source/d3d8to9_base.cpp
@@ -14,7 +14,7 @@ static const D3DFORMAT AdapterFormats[] = {
 };
 
 IDirect3DDevice9 *pCurrentDeviceInterface = nullptr;
-D3DPRESENT_PARAMETERS *pCurrentPresentParams = nullptr;
+D3DPRESENT_PARAMETERS CurrentPresentParams;
 
 // IDirect3D8
 Direct3D8::Direct3D8(IDirect3D9 *ProxyInterface) :
@@ -206,9 +206,9 @@ HRESULT STDMETHODCALLTYPE Direct3D8::CreateDevice(UINT Adapter, D3DDEVTYPE Devic
 
 	HRESULT hr = ProxyInterface->CreateDevice(Adapter, DeviceType, hFocusWindow, BehaviorFlags, &PresentParams, &DeviceInterface);
 
-	if (hr == D3DERR_DEVICELOST && pCurrentDeviceInterface && pCurrentPresentParams)
+	if (hr == D3DERR_DEVICELOST && pCurrentDeviceInterface)
 	{
-		pCurrentDeviceInterface->Reset(pCurrentPresentParams);
+		pCurrentDeviceInterface->Reset(&CurrentPresentParams);
 		hr = ProxyInterface->CreateDevice(Adapter, DeviceType, hFocusWindow, BehaviorFlags, &PresentParams, &DeviceInterface);
 	}
 
@@ -218,7 +218,7 @@ HRESULT STDMETHODCALLTYPE Direct3D8::CreateDevice(UINT Adapter, D3DDEVTYPE Devic
 	}
 
 	pCurrentDeviceInterface = DeviceInterface;
-	pCurrentPresentParams = &PresentParams;
+	CopyMemory(&PresentParams, &CurrentPresentParams, sizeof(PresentParams));
 	*ppReturnedDeviceInterface = new Direct3DDevice8(this, DeviceInterface, (PresentParams.Flags & D3DPRESENTFLAG_DISCARD_DEPTHSTENCIL) != 0);
 
 	return D3D_OK;
--- a/source/d3d8to9_base.cpp
+++ b/source/d3d8to9_base.cpp
@@ -206,7 +206,7 @@ HRESULT STDMETHODCALLTYPE Direct3D8::CreateDevice(UINT Adapter, D3DDEVTYPE Devic
 
 	HRESULT hr = ProxyInterface->CreateDevice(Adapter, DeviceType, hFocusWindow, BehaviorFlags, &PresentParams, &DeviceInterface);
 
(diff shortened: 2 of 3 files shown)


elishacloud:
I forgot to make a new branch so it looks like the check-in to fix [Silent Hill 2](https://github.com/crosire/d3d8to9/issues/24) got put into this pull request.  Basically this undoes part of the pixel shader change I made in [#32](https://github.com/crosire/d3d8to9/pull/32/commits/5d3fb1a53c001e883067c0dcc0bfab90168d442e#diff-366bc9823b5fedd9a67231f1896c4f8fR1767) to fix lighting issues.  I was a bit too zealous when I made that change.

5. Removed regex that changes a '_bx2' modifier for constant values to use _x2 modifier

This change simply removes one line.  I already tested this with other games including _Star Wars Republic Commando_ to make sure I did not revert any of the previous fixes.  This line is not needed, as far as I can tell, for any of the games I have tested.

source/d3d8to9_base.cpp
```diff
@@ -10,10 +10,11 @@ static const D3DFORMAT AdapterFormats[] = {
 	D3DFMT_X8R8G8B8,
 	D3DFMT_R5G6B5,
 	D3DFMT_X1R5G5B5,
-	D3DFMT_A1R5G5B5,
-	D3DFMT_A2R10G10B10
+	D3DFMT_A1R5G5B5
 };
 
+IDirect3DDevice9 *pCurrentDeviceInterface = nullptr;
```
crosire: Shouldn't this be a field of the `Direct3D8` class instead? I'm not sure if it is possible to create different D3D8 instances via `Direct3DCreate8`, but if it was, this could cause race conditions otherwise (very unlikely to happen in the wild, but just theoretically).

CookiePLMonster:
I wonder, should you consider this a bug in GTA III/GTA VC? If so, might be a good idea to fix it in [http://gtaforums.com/topic/669045-silentpatch/](url)

source/d3d8to9_base.cpp
CookiePLMonster: This should be fairly easy to check via a simple unit test, I guess? As it's COM, you probably can't have multiple pointers to the same class and thus if two Direct3DCreate8 calls return the same pointer, then it means it's the same device.

elishacloud: @crosire, Yes, I tried using `Direct3D8` class here first.  It still [crashed](https://github.com/crosire/d3d8to9/issues/39#issuecomment-322025422) after this because GTA VC would call `Release()` and release the old device after we had already released it.  If we release the `Direct3D8` device we will delete it see [here](https://github.com/crosire/d3d8to9/blob/fc2dcac3531bcda61f48dcf0566eb04559832370/source/d3d8to9_base.cpp#L82) which causes GTA VC to crash.  However the `IDirect3DDevice9` pointer is always available even after it is released so GTA VC will not crash if we release `IDirect3DDevice9` but will crash if we release `Direct3D8`.  I also tested both ways and was able to reproduce the crash (after buying the game from Steam).  This may be an issue only with the Steam version because when I used a hacked version (so I did not have to buy the game) it did not crash here.

@CookiePLMonster, Yes, I think this is a bug in GTA.  That is why I commented above and said "GTA does not know how to properly handle this error and simply crashed."

BTW: I wonder if there is a way to do the same thing that Direct3D 9 does here and keep the pointer available even after we delete the class?  This way the game would not crash and we would not need to do caching like we did for [Need for Speed III](https://github.com/crosire/d3d8to9/pull/20).

elishacloud:
I updated this to `Reset` the device rather than releasing it.  This solves the issue and I think is the proper fix.  See comments from Microsoft [here](https://msdn.microsoft.com/en-us/library/windows/desktop/bb174714(v=vs.85).aspx):

> Reset is the only method that has an effect when a device is lost, and is the only method by which an application can change the device from a lost to an operational state.

I also fixed the way I check for `D3DERR_DEVICELOST`.  The previous way was not very good.

source/d3d8to9_base.cpp
```diff
@@ -202,13 +204,21 @@ HRESULT STDMETHODCALLTYPE Direct3D8::CreateDevice(UINT Adapter, D3DDEVTYPE Devic
...
+		hr = ProxyInterface->CreateDevice(Adapter, DeviceType, hFocusWindow, BehaviorFlags, &PresentParams, &DeviceInterface);
+	}
 
 	if (FAILED(hr))
 	{
 		return hr;
 	}
 
+	pCurrentDeviceInterface = DeviceInterface;
+	pCurrentPresentParams = &PresentParams;
```
crosire: That pointer is invalid as soon as the scope is left?!

source/d3d8to9.cpp
```diff
@@ -18,7 +18,10 @@ std::ofstream LOG;
...
 {
 #ifndef D3D8TO9NOLOG
-	LOG.open("d3d8.log", std::ios::trunc);
+	if (!LOG.is_open())
+	{
+		LOG.open("d3d8.log", std::ios::trunc);
+	}
 
 	if (!LOG.is_open())
 	{
```
HerMajestyDrMona: Maybe it's good to add an additional bool, so it won't spam MessageBoxes, every time user change the screen resolution and the log fails to open?

crosire: Changing the screen resolution does not require recreation of the entire D3D8 COM interface. I'd be surprised if any game calls this function at any other time other than startup (it may call it multiple times there, but it's unlikely to happen again once it's running).

HerMajestyDrMona: You're right, sorry. i thought that this function was related to those GTA issues reported earlier. Anyway if in some cases this function is called multiple times on startup, I think it's good to limit displaying of the message box to one time.

elishacloud:
Thanks for the all the comments.  I removed pointer from PresentParams so that the data would not be lost when the scope is lost (though I don't I did not see any issue with it from my tests) and added a flag to the logging so that message box only is shown once.

Let me know if there are any other changes that are needed here.

source/d3d8to9_base.cpp
```diff
@@ -202,13 +204,21 @@ HRESULT STDMETHODCALLTYPE Direct3D8::CreateDevice(UINT Adapter, D3DDEVTYPE Devic
 
 	IDirect3DDevice9 *DeviceInterface = nullptr;
 
-	const HRESULT hr = ProxyInterface->CreateDevice(Adapter, DeviceType, hFocusWindow, BehaviorFlags, &PresentParams, &DeviceInterface);
+	HRESULT hr = ProxyInterface->CreateDevice(Adapter, DeviceType, hFocusWindow, BehaviorFlags, &PresentParams, &DeviceInterface);
+
+	if (hr == D3DERR_DEVICELOST && pCurrentDeviceInterface)
+	{
+		pCurrentDeviceInterface->Reset(&CurrentPresentParams);
```
crosire: Just wondering, but shouldn't it also work to reset the current device with the present parameters of the new device? That way we could avoid the second global variable.

elishacloud: Great idea!  But I just tried it and I see a crash sometimes when changing resolution if I do that.  Not sure why but it could be related to either the `hDeviceWindow` or maybe even one of the `D3DFORMAT` parameters (such as `BackBufferFormat` or `AutoDepthStencilFormat`).  Too bad.

crosire: Alright =)

PR #46, opened 2017-09-10
Update Ref Count

elishacloud:
This update will fix a couple of issues with the reference counter.  This fixes the crash in [GTA III/VC](https://github.com/crosire/d3d8to9/issues/39) when changing resolutions and the crash in [Serious Sam 1 & 2](https://github.com/crosire/d3d8to9/issues/39#issuecomment-323473718) when pressing alt+tab.

There are several fixes in this update:

1. This update stops adding a reference the constructor and releasing the reference in each destructor.

These extra references are not needed.  Plus they complicate the code and masks other issues (see below).  Also it seems like d3d8to9 does not always close out all the references here, though I cannot prove this.  This seems to be the main issue because the reason the games were crashing is because the previous device was still open and the reason the previous device was still open was because there were extra references on it.

2. This update stops releasing a reference after calling `GetType()` in the `GetTexture()` function.

This extra `Release()` would have been causing a crash in many games but since we added the extra reference in the constructor (see above) this issue was being masked.  The documentation [here](https://msdn.microsoft.com/en-us/library/windows/desktop/bb205883(v=vs.85).aspx), [here](https://msdn.microsoft.com/en-us/library/windows/desktop/bb205876(v=vs.85).aspx) and everywhere else I have looked says nothing about `GetType()` adding a reference counter.  From my testing `GetType()` never adds a reference counter.  So we should not have a `Release()` in the `GetTexture()` function.

3. This update no longer removes the object from the address cache in the destructor.

All the objects are removed only when the device is removed, to mirror Direct3D9's behavior.  Therefore we don't need to remove the cache because it will already get removed by the AddressLookupTable destructor [here](https://github.com/crosire/d3d8to9/blob/master/source/lookup_table.cpp#L24).

4. This update removes the previous workaround for the lost device see [here](https://github.com/crosire/d3d8to9/commit/436c84bae9a68b82f54cb22b5830d272e321c74d)

We no longer need that extra code for resetting the device now that the reference counter is corrected.  That was really only fixing the symptom not the root issue.

### DIFF
--- a/source/d3d8to9_base.cpp
+++ b/source/d3d8to9_base.cpp
@@ -13,9 +13,6 @@ static const D3DFORMAT AdapterFormats[] = {
 	D3DFMT_A1R5G5B5
 };
 
-IDirect3DDevice9 *pCurrentDeviceInterface = nullptr;
-D3DPRESENT_PARAMETERS CurrentPresentParams;
-
 // IDirect3D8
 Direct3D8::Direct3D8(IDirect3D9 *ProxyInterface) :
 	ProxyInterface(ProxyInterface)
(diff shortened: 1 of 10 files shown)


crosire:
Awesome!

PR #48, opened 2017-09-16
Address cache

elishacloud:
Simplify cache by moving to unordered_map.  This removes the for loops in the `FindAddress()` function and slightly simplifies the rest of the code.


source/lookup_table.hpp
```diff
@@ -98,21 +79,19 @@ class AddressLookupTable
 		}
 
 		constexpr UINT CacheIndex = AddressCacheIndex<T>::CacheIndex;
-
-		for (size_t i = 0; i < AddressCache[CacheIndex].size(); i++)
+		for (auto it = AddressCache[CacheIndex].begin(); it != AddressCache[CacheIndex].end(); ++it)
```
CookiePLMonster: Fairly sure this could be replaced with `std::find_if` and an appropriate predicate.

elishacloud: Good idea!  Updated code to use find_if rather than a for loop.

source/lookup_table.hpp
```diff
@@ -56,16 +49,7 @@ class AddressLookupTable
...
-		constexpr UINT CacheIndex = AddressCacheIndex<T>::CacheIndex;
-
-		for (size_t i = 0; i < AddressCache[CacheIndex].size(); i++)
-		{
-			if (AddressCache[CacheIndex][i].Address9 == pAddress9)
-			{
-				pAddress8 = static_cast<T *>(AddressCache[CacheIndex][i].Address8);
-			}
-		}
+		T *pAddress8 = static_cast<T *>(AddressCache[AddressCacheIndex<T>::CacheIndex][pAddress9]);
```
CookiePLMonster: I haven't noticed this earlier, but this changes behaviour of the function. Old variation didn't modify the container if address wasn't found. This one will push a `nullptr` entry for `pAddress9` key if it doesn't exist. Might be harmless, but might not have been intentional - it's probably safer to use `.find()`.

Also, unrelated but this code seems to be locating elements by both key and value. I'm not sure what's @crosire's standpoint on using Boost (personally I'm not a fan), but this use case seems to be perfect for Boost.Bimap:
 [http://www.boost.org/doc/libs/1_65_1/libs/bimap/doc/html/index.html](url)

crosire: @CookiePLMonster I'd rather not add additional third-party dependencies, especially for simple things like a map.

CookiePLMonster: Fair enough. I really doubt this would be a performance bottleneck anyway, but if it ever becomes one then two synced maps will do just fine.

elishacloud: @CookiePLMonster, Thanks for the comments!  But, are you sure about this?  Is it changing `pAddress9` to `nullptr` or `pAddress8` to `nullptr`?  It is suppose to change `pAddress8` to `nullptr`.

I tried changing this:
```c++
T *pAddress8 = static_cast<T *>(AddressCache[AddressCacheIndex<T>::CacheIndex][pAddress9]);
```

to this:

```c++
T *pAddress8 = static_cast<T *>(AddressCache[AddressCacheIndex<T>::CacheIndex].find(pAddress9)->second);
```

and it causes issues in a couple of the games I tested.  For example in _Need for Speed III_ it would cause the game to crash.

What I had first seemed to work fine for me.  I have had it that way for a while and tested a number of games without any issues.  I am also using this for other [wrappers](https://github.com/elishacloud/DirectX-Wrappers) I am creating and it has worked on every game I tested in DirectX7, Direct3D8 and Direct3D9.  Can you retest it?

CookiePLMonster: If this `pAddress9` doesn't exist, using operator[] on a map creates a key with a default value - in your case, it creates a key with value 0, thus returned `pAddress8` is nullptr **and** it's put in the map that way.

elishacloud: Good catch!  I totally missed that.  Updated change should fix that issue.

source/lookup_table.hpp
```diff
@@ -57,17 +51,13 @@ class AddressLookupTable
 		}
 
 		T *pAddress8 = nullptr;
```
crosire: No need for a temporary variable here: Can just return the result in the branch below.

elishacloud: Updated.  Removed temporary variable.

Issue #70, opened 2018-04-06
Translation error when moving _bx2 to separate register


PatrickvL:
Pixel shader input :

	dp3 r0.rgb, c0_bx2, t0_bx2

d3d8to9 output :

	mov r0.xyz, c0 /* added line */
	dp3 r0.xyz, r0.xyz_bx2 /* changed c0 to r0.xyz */, t0_bx2

This fails with :

	> Failed to reassemble shader :

	error X2004 : invalid swizzle 'xyz_bx2'
	error X5027 : Invalid src swizzle for first source param.

elishacloud:
The issue is with [this](https://github.com/crosire/d3d8to9/blob/9f1d3670db20269f2bf1a31e4431266c9d121de2/source/d3d8to9_device.cpp#L1783-L1785) line and [this](https://github.com/crosire/d3d8to9/blob/9f1d3670db20269f2bf1a31e4431266c9d121de2/source/d3d8to9_device.cpp#L1789-L1791) line.  I purposely kept the swizzle to try and preserve the original pixel shader functionality.  Removing the swizzle has a possibility of changing the final register value, which could cause an issue.

It is pretty easy to remove the swizzle here, but I would need to test many games to ensure that it does not cause any issues.  It might be better if I remove the swizzle only in the case where there is a rear [modifier](https://msdn.microsoft.com/en-us/library/windows/desktop/bb219868(v=vs.85).aspx), like `_bx2` or `_bias`.  However that would require duplicating those lines, which I really don't want to do.

_Note: you cannot have both a rear modifier and a swizzle._

elishacloud:
As I thought, simply removing the swizzle caused issues with _True Crime New York City_.  However I forgot that you can use a rear modifier with a swizzle as long as the swizzle comes last.  So it should look like this:

d3d8to9 output :

```asm
mov r0.xyz, c0 /* added line */
dp3 r0.xyz, r0_bx2.xyz /* changed c0 to r0.xyz */, t0_bx2
```

Try with [this](https://github.com/crosire/d3d8to9/files/1885882/d3d8.zip) file and see if it works.

PatrickvL:
This still fails to compile - I had to change the input shader to use an intermediate register to avoid the swizzle;

This problematic input :
>     dp3 r0.rgb, c0_bx2, t0_bx2

Gets converted to this, which doesn't work :
>     mov r0.xyz, c0 /* added line */
[...]

PR #75, opened 2018-04-14
Remove swizzle from changed lines

elishacloud:
## Overview

This fixes the comments mentioned [here](https://github.com/crosire/d3d8to9/issues/70#issuecomment-381148614) by @PatrickvL. This check-in removes the swizzle from the modified line while keeping the swizzle in the added `mov` line.

Since it keeps the swizzle on the `mov` line it still allows for the closest possible conversion.

However, it needs to remove the swizzle from the source modifier because not all source modifiers can be used together with a swizzle.  Note: a swizzle is just another form of a source modifier.

See documentation [here](https://msdn.microsoft.com/en-us/library/windows/desktop/bb219868(v=vs.85).aspx):

> * Negate can be combined with either the bias, signed scaling, or scalex2 modifier. When combined, negate is run last.
> * Invert cannot be combined with any other modifier.
> * Invert, negate, bias, signed scaling, and scalex2 can be combined with any of the selectors.
> * Source register modifiers should not be used on constant registers because they will cause undefined results. For version 1_4, modifiers on constants are not allowed and will fail validation.

Below are two examples of how the new translations look with this check-in.

**Example 1:**
So this:
```asm
dp3 r0.rgb, c0_bx2, t0_bx2
```

Gets changed to:
```asm
mov r0.xyz, c0 /* added line */
dp3 r0.xyz, r0_bx2 /* changed c0 to r0 */, t0_bx2
```

**Example 2:**
This:
```asm
    lrp r0.xyz, c0.w, r1, r0
  + mul r0.w, 1-v0.w, 1-c0.w
```

Gets changed to:
```asm
    mov r0.w, c0.w /* added line */
    lrp r0.xyz, c0.w, r1, r0
  + mul r0.w, 1-v0.w, 1-r0 /* changed c0.w to r0 */
```

## Testing: 

Tested with:

* Silent Hill 2
* Star Wars Republic Commando
* TOCA Race Driver 2
* True Crime New York City


crosire:
:+1:

Issue #42, opened 2017-08-24
 [Silent Hill 2] red tinted cutscenes


scorutw:
heyo.
recently beat first silent hill, hell of a job to emulate, but now im having trouble setting up the PC director's cut with the widescreen fix
every cutscene ingame looks like this
![29591878-c32aab30-8766-11e7-8153-c1f1cd3749b8-1](https://user-images.githubusercontent.com/31296305/29644372-22daee58-883b-11e7-821e-6a83c6dd61f7.png)
its pretty annoying and i havent seem to find any success in fixing it.
i think it might have something to do with the fact that i use integrated graphics but id like to know if there's a way to combat this and have it working properly with the widescreen fix.
i looked through this thread but didnt find anything
https://github.com/crosire/d3d8to9/issues/24
thanks

elishacloud:
Yes, this seems to be an issue that only happens when using d3d8to9 and an Intel GPU.  I believe it is a problem with d3d8to9 but I have not been able to find the issue yet.

As far as a work around you can use [WineD3D](https://fdossena.com/?p=wined3d/index.frag) to fix this issue.  There are two options for you:

1. If you are using the [Widescreen Fixes Pack](https://thirteenag.github.io/wfp#sh2) for Silent Hill 2 you can use the [LoadWineD3D ASI](https://github.com/elishacloud/LoadWineD3D-ASI) that I created.  Simply copy the LoadWineD3D.asi + d3d9.dll + wine3d.dll + libwine.dll into the 'scripts' folder.

2. Otherwise you can simply copy the WineD3D files (d3d9.dll + wine3d.dll + libwine.dll) into the Silent Hill 2 folder.

Also take a look at the [Silent Hill 2 Fix Page](http://widescreenfix.townofsilenthill.com/SH2/). This has a bunch of stuff for Silent Hill 2 that will make the experience better.

I hope to eventually put a fix into d3d8to9 for this red tint issue but using WineD3D completely solves the issue for now.

scorutw:
hey i tried that loadwined3d asi fix but it didnt work :(
this is my silent hill /scripts/ folder
![9d7d2f9a3c9da3a09d55b5fe22a3badf](https://user-images.githubusercontent.com/31296305/29738305-cfd547a2-89e4-11e7-811d-cc4de71e978d.png)
this is my regular outfront silent hill 2 folder
[...]

PR #77, opened 2018-05-03
Disable YUV formats on textures

elishacloud:
This pull request fixes issue #42.

## Overview

This change will return `D3DERR_NOTAVAILABLE` when an application queries using `CheckDeviceFormat` on any texture using a YUV format.  ATI and Nvidia already return this value for textures, so it should have no affect on these GPUs.  However, some Intel chips do not.  It appears that these Intel chips attempt to handle the formats but do not handle them correctly.

I believe with Direct3D9 that YUV is supposed to be unsupported on textures.  See [comment](https://www.winehq.org/pipermail/wine-patches/2014-February/130383.html) on WineD3D:

>         /* Some (all?) Windows drivers do not support YUV 3D textures, only 2D surfaces in
>          * StretchRect. Thus use StretchRect to draw the YUV surface onto the screen instead
>          * of drawPrimitive. */

## Notes

I did think about implementing full support for these formats into d3d8to9, but since most video cards don't support them I did not see the need to do that.

I also thought about blocking the creation of textures that use these formats.  I did some tests on Nvidia and it blocks the creation of a texture that use these formats.  However this issue was fixed by simply telling the application they are unsupported so I did not add any block on the creation of textures with these formats.

We can always add more fixes here later if there is an issue with other games.

## Testing

I tested with the following games:

 - Codemasters Race Driver 2
 - Grand Theft Auto 3
 - Grand Theft Auto Vice City
 - Haegemonia Legions of Iron
 - Haegemonia The Solon Heritage
 - Hitman 2 Silent Assassin
 - Legacy of Kain Blood Omen 2
 - Max Payne
 - Max Payne 2
 - Need for Speed III
 - Need For Speed Hot Pursuit 2
 - Raymond 3
 - Serious Sam The First Encounter
 - Serious Sam The Second Encounter
 - Silent Hill 2
 - Star Wars Republic Commando
 - Star Wars Starfighter
 - True Crime New York City

### DIFF
--- a/source/d3d8to9_base.cpp
+++ b/source/d3d8to9_base.cpp
@@ -145,8 +145,7 @@ HRESULT STDMETHODCALLTYPE Direct3D8::CheckDeviceType(UINT Adapter, D3DDEVTYPE Ch
 }
 HRESULT STDMETHODCALLTYPE Direct3D8::CheckDeviceFormat(UINT Adapter, D3DDEVTYPE DeviceType, D3DFORMAT AdapterFormat, DWORD Usage, D3DRESOURCETYPE RType, D3DFORMAT CheckFormat)
 {
-	if ((RType == D3DRTYPE_TEXTURE || RType == D3DRTYPE_VOLUMETEXTURE || RType == D3DRTYPE_CUBETEXTURE) &&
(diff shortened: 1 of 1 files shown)


crosire:
I don't think the additional check to see if `RType` is a texture is necessary. A YUV vertex/index buffer doesn't make sense, so might as well always return the error for those formats, regardless of all other arguments.

elishacloud:
Yes, good point.  Somehow I missed that!  Updated in [e4a0e32](https://github.com/crosire/d3d8to9/pull/77/commits/e4a0e3230301c425c68c9d5f5289edd1f0652142).

PR #78, opened 2018-05-15
Skip unneeded SetViewport calls

### LINKED ISSUE #76: Hydorah improper window sizing?
So a free game by Locomalito - Hydorah, does not size the game's window properly when using d3d8.dll + Reshade's d3d9.dll.  What happens is only a quarter of the screen is visible in the game window--zoomed way in, while the rest is cut off and you can't see it.

Very strange.

It shouldn't be doing this?  The game supports multiple screen sizing modes, and every single one of them behaves the same way, no matter the window size, be it small or fullscreen.

Remove d3d8.dll and reshade, everything is fine again.  Hydorah is dx 8.

elishacloud:
## Overview:

This pull fixes #76.  It will skip unneeded calls to `SetViewport`.

The extra call to `SetViewport` causes an issue in Direct3D9, while Direct3D8 can handle it.  I believe this is a bug in Hydorah because based on my testing even in Direct3D8 this call is not needed.  But since Direct3D9 does not handle it the same as Direct3D8 I believe we should fix it.

**This update makes the following changes:**
If `SetViewport` is called twice in the same frame and the only significant function called in between these two calls is `SetTransform` using `D3DTS_VIEW` it will skip the second call to `SetViewport`.

Significant calls are defined as:
* BeginScene()
* CopyRects()
* DrawIndexedPrimitive()
* DrawIndexedPrimitiveUP()
* DrawPrimitive()
* DrawPrimitiveUP()
* DrawRectPatch()
* DrawTriPatch()
* LightEnable()
* Present()
* ProcessVertices()
* Reset()
* SetIndices()
* SetLight()
* SetMaterial()
* SetPixelShader()
* SetRenderTarget()
* SetRenderState()
* SetStreamSource()
* SetTexture()
* SetTextureStageState()
* SetVertexShader()
* UpdateTexture()

Note: Hydorah calls the following three functions in this order:
```c++
SetViewport(0x0019FBCC);
SetTransform(D3DTS_VIEW, 0x0019FB80);
SetTransform(D3DTS_PROJECTION, 0x0019FB80);
Clear(0x00000000, NULL, 0x00000003, D3DCOLOR_ARGB(0xff,0x00,0x00,0x00), 1.000f, 0x00000000);
SetViewport(0x0019FBCC);        // <-- This one is not needed and gets skipped
```

## Testing:

I tested this with the following games:

* Codemasters Race Driver 2
* Grand Theft Auto 3
* Grand Theft Auto Vice City
* Haegemonia Legions of Iron
* Haegemonia The Solon Heritage
* Hitman 2 Silent Assassin
* Hydorah
* Indiana Jones and the Emperor's Tomb
* Legacy of Kain Blood Omen 2
* Max Payne
* Max Payne 2
* Moto Racer 3
* Need for Speed III
* Need For Speed Hot Pursuit 2
* Raymond 3
* Serious Sam The First Encounter
* Serious Sam The Second Encounter
* Silent Hill 2
* Star Wars Republic Commando
* Star Wars Starfighter
* True Crime New York City

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -191,6 +191,8 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::Reset(D3DPRESENT_PARAMETERS8 *pPresen
 		return D3DERR_INVALIDCALL;
 	}
 
+	NewScene = true;
+
 	D3DPRESENT_PARAMETERS PresentParams;
 	ConvertPresentParameters(*pPresentationParameters, PresentParams);
 
@@ -217,6 +219,8 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::Reset(D3DPRESENT_PARAMETERS8 *pPresen
 HRESULT STDMETHODCALLTYPE Direct3DDevice8::Present(const RECT *pSourceRect, const RECT *pDestRect, HWND hDestWindowOverride, const RGNDATA *pDirtyRegion)
 {
 	UNREFERENCED_PARAMETER(pDirtyRegion);
+
+	NewScene = true;
 
 	return ProxyInterface->Present(pSourceRect, pDestRect, hDestWindowOverride, nullptr);
 }
@@ -494,6 +498,8 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::CopyRects(Direct3DSurface8 *pSourceSu
 		return D3DERR_INVALIDCALL;
 	}
 
+	NewScene = true;
+
 	D3DSURFACE_DESC SourceDesc, DestinationDesc;
 	pSourceSurface->GetProxyInterface()->GetDesc(&SourceDesc);
 	pDestinationSurface->GetProxyInterface()->GetDesc(&DestinationDesc);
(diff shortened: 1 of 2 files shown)


source/d3d8to9_device.cpp
```diff
@@ -705,6 +722,14 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::MultiplyTransform(D3DTRANSFORMSTATETY
 }
 HRESULT STDMETHODCALLTYPE Direct3DDevice8::SetViewport(const D3DVIEWPORT8 *pViewport)
 {
+	if (ViewpointFlag && NewScene == false)
```
crosire: I'm not a fan of the `== false` syntax. Would you mind changing this to `!NewScene`?

crosire:
This is a weird bug. The only way I can explain this to myself is that `SetViewport` is affected by whatever view transform is currently set. In that case the cleaner solution would be to set the transform to the identity matrix before applying the viewport and then resetting it to whatever it was before. But I don't know if that's it.

elishacloud:
**Hold off on this pull request.**  I think I found the root cause.  It looks like the call is failing in Direct3D8 because the parameters in `SetViewport` are wrong because the width and height are too large.  As far as I can tell, Direct3D9 is also suppose to fail this call, but does not.  I am working on a better fix for this but need to do more testing.

According to the documentation [here](https://msdn.microsoft.com/en-us/library/windows/desktop/bb174469(v=vs.85).aspx) it should "return D3DERR_INVALIDCALL ... if pViewport is invalid, or if pViewport describes a region that cannot exist within the render target surface."

It appears that Direct3D9 is still accepting the call even though the "region that cannot exist within the render target surface."

Issue #61, opened 2018-03-01
Problem playing .bik movie files ingame while using d3d8to9


Radio30:
This in Indiana Jones and the Emporer's Tomb.

All the intro and cut-scene movies are in .bik format and while using d3d8to9 the audio plays during the videos but the screen is black.
Any idea what could be causing this?

elishacloud:
Yes this issue was introduced with Windows 10 [Fall Creators Edition](https://blogs.windows.com/windowsexperience/2017/10/17/whats-new-windows-10-fall-creators-update/).  If you are running any older version of Windows 10 (or other versions of Windows) this issue does not happen.  I noticed this a while ago and was planning to file an issue on this, but you saved me the trouble.

I have seen this "black screen" issue on a number of games starting with the [Creators Edition](https://blogs.windows.com/windowsexperience/2017/04/11/whats-new-in-the-windows-10-creators-update/#tLCYX7W76LVRXEEE.97).  See [here](https://github.com/narzoul/DDrawCompat/issues/15) for a similar example.

HerMajestyDrMona:
Just information: I'm using Windows 10 Fall Creators Update (16299.248 currently) and .bik movies work fine in Empire Earth 2. They work with and without d3d8to9.

elishacloud:
Yes, I believe the issue is specific to _Indiana Jones and the Emporer's Tomb_ and not directly related to .bik movies.

wetcoder:
**[Indiana Jones and The Emperor's Tomb (2003) Full controller support FIX  + Widescreen Fix](https://www.gog.com/forum/indiana_jones_series/controller_support_for_emperors_tomb/post7)**

elishacloud:
I just found out that _Indiana Jones and the Emperor's Tomb_ uses [DirectDraw](https://msdn.microsoft.com/en-us/library/windows/desktop/gg426115(v=vs.85).aspx).  I noticed it calling:

- [DirectDrawCreate](https://msdn.microsoft.com/en-us/library/windows/desktop/gg426116(v=vs.85).aspx)
- [SetCooperativeLevel](https://msdn.microsoft.com/en-us/library/windows/desktop/gg426155(v=vs.85).aspx)
- [CreateSurface](https://msdn.microsoft.com/en-us/library/windows/desktop/gg426136(v=vs.85).aspx)

However `CreateSurface` fails with error 887600E1.  This happens with or without d3d8to9.

[...]

PR #108, opened 2020-02-03
Fix vsync, oPos register and RenderState

### LINKED ISSUE #49: Graphical and crashing issues on FireStarter 2004 by gsc game world
yknow those stalker guys. Its £5 on gog https://www.gog.com/game/firestarter

issues such as randomly stretched to horizon polygons, flickering out of existance monitors and just straight up crashes to desktop.

### LINKED ISSUE #63: Ive asked before. But is there no way we can find out why AA doesn't work with the module?
The games i play like serious sam benefit so heavily from aa that I really cannot go back just for more frames. Is there no way to log what is blocking it?

### LINKED ISSUE #86: Tony Hawks Pro Skater 3 blue background
note: congratulations for this great work, this same dll improvement graphics with the emulator of xbox classic would be great improvements to make more compatible graphic errors in the emulator, the good thing about this dll is that I can use the reshader in some games that did not work

blue background
image

without d3d8to9
image

### LINKED ISSUE #103: How to make nvidia's cp 3d settings work with d3d8to9?
Hello,

None of the 3d settings from the cp panel seem to work while using d3d8to9, like AA, AF etc. Is there any way? The game I play is cnc generals btw.

elishacloud:
This pull request fixes #49 
This pull request fixes #63 
This pull request fixes #86 
This pull request fixes #103 
This pull also fixes [AquaNox](https://www.gog.com/game/aquanox), which never worked before with d3d8to9.

There are three main updates in this pull:

1. Fixes present parameters to match the documentation.

This will properly set the vsync based on whether the game is usin windowed or fullscreen.  This also removes the code that forced `D3DMULTISAMPLE_NONE`, as this broke some AA functions.

2. Fixes oPos register swizzles in VertexShader

AquaNox uses almost 100 VirtexShaders.  Several of them would set only  on the x and y components of the pPos register, which is not allowed in DirectX9.  See sample below.

Before fix (see error):
```asm
> Dumping translated shader assembly:

    vs_1_1
    dcl_position v0
    dcl_blendweight v1
    dcl_blendindices v2
    dcl_normal v3
    dcl_psize v4
    mov r0, c0 /* initialize register r0 */
    mov oT1, c0 /* initialize output register oT1 */
    mov oT0, c0 /* initialize output register oT0 */
    mul r0.xy, v3, c23
    add oPos.xy, r0, c22      //  <--  Error is here
    mov oT0.xy, v3
    mul r0.xy, v3, c25
    add oT1.xy, r0, c24

// approximately 5 instruction slots used

> Failed to reassemble shader:

d:\games\aquanox\memory(17,1): error X5350: Vertex shader must minimally write all four components (xyzw) of oPos output register.  Missing components(*): x/0 y/1 *z/2 *w/3
```

After fix:

```asm
> Dumping translated shader assembly:

    vs_1_1
    dcl_position v0
    dcl_blendweight v1
    dcl_blendindices v2
    dcl_normal v3
    dcl_psize v4
    mov r0, c0 /* initialize register r0 */
    mov oT1, c0 /* initialize output register oT1 */
    mov oT0, c0 /* initialize output register oT0 */
    mul r0.xy, v3, c23
    add oPos, r0, c22 /* removed oPos swizzles */
    mov oT0.xy, v3
    mul r0.xy, v3, c25
    add oT1.xy, r0, c24

// approximately 5 instruction slots used
```

3. Changes how SetRenderState/GetRenderState works with the following States: `D3DRS_LINEPATTERN`, `D3DRS_EDGEANTIALIAS` and `D3DRS_PATCHSEGMENTS`

These states were removed in DirectX9.  For `D3DRS_LINEPATTERN` and `D3DRS_PATCHSEGMENTS` this changes from returning `D3DERR_INVALIDCALL` to returning `D3D_OK`.  It also changes `D3DRS_EDGEANTIALIAS` to use `D3DRS_ANTIALIASEDLINEENABLE` instead.

These changes are required for AquaNox because AquaNox would monitor the return call and fail out if anything other than `D3D_OK` is returned.

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -787,13 +787,14 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::SetRenderState(D3DRENDERSTATETYPE Sta
 
 	switch (static_cast<DWORD>(State))
 	{
-		case D3DRS_LINEPATTERN:
 		case D3DRS_ZVISIBLE:
-		case D3DRS_EDGEANTIALIAS:
-		case D3DRS_PATCHSEGMENTS:
 			return D3DERR_INVALIDCALL;
+		case D3DRS_PATCHSEGMENTS:
+		case D3DRS_LINEPATTERN:
 		case D3DRS_SOFTWAREVERTEXPROCESSING:
 			return D3D_OK;
+		case D3DRS_EDGEANTIALIAS:
+			return ProxyInterface->SetRenderState(D3DRS_ANTIALIASEDLINEENABLE, Value);
 		case D3DRS_CLIPPLANEENABLE:
 			hr = ProxyInterface->SetRenderState(State, Value);
 			if (SUCCEEDED(hr))
@@ -821,10 +822,13 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::GetRenderState(D3DRENDERSTATETYPE Sta
 
 	switch (static_cast<DWORD>(State))
 	{
-		case D3DRS_LINEPATTERN:
 		case D3DRS_ZVISIBLE:
-		case D3DRS_EDGEANTIALIAS:
 			return D3DERR_INVALIDCALL;
+		case D3DRS_LINEPATTERN:
+			*pValue = 0;
+			return D3D_OK;
+		case D3DRS_EDGEANTIALIAS:
+			return ProxyInterface->GetRenderState(D3DRS_ANTIALIASEDLINEENABLE, pValue);
 		case D3DRS_ZBIAS:
 			hr = ProxyInterface->GetRenderState(D3DRS_DEPTHBIAS, pValue);
 			*pValue = static_cast<DWORD>(*reinterpret_cast<const FLOAT *>(pValue) * -500000.0f);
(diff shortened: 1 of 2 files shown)


crosire:
Awesome work!

CookiePLMonster:
This makes GTA III and Vice City not use vsync when running in windowed mode (forced by another mod), even though games should request vsync. Is this intended?

CookiePLMonster:
> I found out if you call `Direct3D9SetSwapEffectUpgradeShim(0)` it will switch the Direct3D9 mode back.

Would that be worth doing for d3d8to9 then, or is forced-no-vsync the behaviour you desire in this scenario?

elishacloud:
> > I found out if you call `Direct3D9SetSwapEffectUpgradeShim(0)` it will switch the Direct3D9 mode back.
> 
> Would that be worth doing for d3d8to9 then, or is forced-no-vsync the behaviour you desire in this scenario?

See comments [here](https://github.com/crosire/d3d8to9/issues/61#issuecomment-1847286903) to answer that question.

mirh:
Ok so, since #61 is already heavy and "focused" on another more practical matter, here I wanted to discuss the question of whether KGL may not just be a "dumb list" but if it couldn't be actually shipping some kinds of knobs itself.

Because, on the one hand the gta games of course *had* to be present in such a big and comprehensive list (there's even an entry for `gta_sa_compact.exe` and `gta vice ultimate trainer.exe` lol).
On the other, it would seem strange that they'd be included into anything considering the older ones aren't even dx9?

---
The following happens inside of dxgi.dll (used for briefness sorry, d3d9 is too much of a [matryoshka](https://github.com/narzoul/DDrawCompat/issues/269#issuecomment-2562065118) of worms).
`CDXGIFactory::FinalConstruct` (whatever that may be, I can guess it's the last step before setting up the swapchain maybe?) has a call to `QueryMaximizedWindowedFullscreen`.

That contains a call for function `QueryUserGlobalSettings`* and it has a lot of interesting "compatstrings" on its own right (`DisableMaximizedWindowedFullscreen`, `FullscreenProxyWindowPromotion`, `DisableMaximizedWindowedUpgrades`) but eventually the kingmaker is the invoking of `FindGameConfigFromStoreAndFeatures`.
That in turn has `QueryDxgGameConfigStore` and `QueryDxgGameConfigStoreGPropHonorFSEWinCompat`. 
I loose any lead afterwards, which I can just scapegoat on them starting to use COM magic.

But I'll be damned if that has nothing to do with *Flags* or the `GameDVR_DXGIHonorFSEWindowsCompatible` thing inside of ResourcePolicyServer (which implements *GameConfigStoreServer*, aka GcsSrv, btw).
After that it's all masks and bit operations, so I gave up. Just to pretend to be smarter (and pick up some extra SEO) some functions in there are `AddGameConfig`, `ModifyGameConfig`, `ValidateGameConfig` and `VerifyIsValidAUMID`. 

*which in turn contains `PerAppUserSettingsHelper::QueryUserSettings(parameters, L"DirectXUserGlobalSettings", results)`.
That first queries for `HKCU\SOFTWARE\Microsoft\DirectX\UserGpuPreferences`, and then string "GpuPreference" in a sea of *directxdatabasehelper.cpp* comments. So this is not particularly crazy or unexpected imo, and it's where the stuff I mentioned last year in https://github.com/crosire/d3d8to9/issues/61#issuecomment-1849212674 is implemented (I didn't have any other value presumably just because the reminder was added in W11)

---
I then proceeded to look around the other side of the equation (twinui.dll, which [apparently](https://web.archive.org/web/20230704194310/https://lucasg.github.io/2017/01/07/Renaming-a-Win32-executable-to-main-exe-launch-the-Xbox-DVR-popup/) is a fond user _ResourcePolicyClient_).. and it's an even more opaque rabbit hole of (probably) COM intricacies but I could find these between the symbols:

    UpdateGameConfigStoreFromKGL
    CheckGameConfigStoreForIsGameFromAppId
    CheckGameConfigStoreForIsGameFromProcessId
    GetAutoGameModeFromGameConfigStore
    ScanKGLForFSECompat

And from the fact that I cannot get @lucasg's template to work, I'm kinda suspecting that the database format isn't just a list of stupid names (plus optionally some extra path specifiers) but that there may be a growing number of properties attached to them.

PR #120, opened 2020-12-07
Fix compilation under MinGW-w64

rafalh:
Fix compilation of d3d8to9 under MinGW-w64.
I was testing using MinGW-w64 9.3 (one distributed with Ubuntu 20.04). I was building using CMake&Ninja (I can contribute a simple CMake build file if you find it useful).
Note: I didn't test it under MSVC but I tried to make it compatible to both compilers.


crosire:
Nice! A simple CMake would be neat as well, sure. Probably can be something similar to https://github.com/crosire/blink/blob/main/CMakeLists.txt.

PR #121, opened 2020-12-08
Add CMake build tool support

rafalh:
As a bonus fix some warnings from GCC (wrong order of initialization in constructor) and improve formatting (unnecessary space at the end of the line).
I allowed optional build as a static library, may be useful if someone wants to embed your code into an executable (I was considering it myself).
I also disabled some warnings from GCC because I wasn't sure if it would make sense to fix them (one would require to add `virtual` modifier to two destructors, another was about `#pragma region`, not sure if it can be ifdefed and still work in MSVC - both warnings are not a real bugs so disabling them is okay IMO).
Also I used slightly different style of CMakeFile.txt than in your blink project:
* cmake functions are lower-case because that's how cmake official documentation uses them (and I like it more)
* file list is explicit (no glob) - it is better approach because when pulling changes that involve adding/removing files developer don't have to force CMake regeneration manually


CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+if(D3D8TO9_STATIC)
+  add_library(d3d8to9 STATIC ${SRCS})
+else()
+  add_library(d3d8to9 SHARED ${SRCS})
+  set_target_properties(d3d8to9 PROPERTIES PREFIX "")
+  set_target_properties(d3d8to9 PROPERTIES OUTPUT_NAME "d3d8")
+endif()
+
+target_compile_definitions(d3d8to9 PUBLIC
+  $<$<CONFIG:Debug>:D3D_DEBUG_INFO;_DEBUG;aaa>
```
crosire, line 36: What's up with the `aaa`? I'm guessing this is left-over debugging code?
Btw., doesn't CMake already set `NDEBUG` automatically, so it should not be necessary to include that here? On that note I see there is some use of `_DEBUG` in d3d8to9 which should be changed to `NDEBUG` at some point ... but that's not necessarily in scope for this PR.

source/d3d8to9_texture.cpp
```diff
@@ -281,8 +281,9 @@ HRESULT STDMETHODCALLTYPE Direct3DCubeTexture8::AddDirtyRect(D3DCUBEMAP_FACES Fa
 }
 
 Direct3DVolumeTexture8::Direct3DVolumeTexture8(Direct3DDevice8 *device, IDirect3DVolumeTexture9 *ProxyInterface) :
-	ProxyInterface(ProxyInterface),
-	Device(device)
+	Device(device),
+	ProxyInterface(ProxyInterface)
+
```
crosire, line 286: Could you remove the extra space?

crosire (commented):
Looking good, just two nitpicks:

source/d3d8to9_texture.cpp
rafalh: Fixed

CMakeLists.txt
rafalh: Fixed! I don't know how could I forget about that 'aaa' :P

rafalh: I've seen you readded `_DEBUG` definition back in 654588e408a985e2474cc8ca82e27f7854b82530 . Is there any reason for that? You also didn't make analogous change in CMakeBuild.txt

crosire: As it was CMake did not add the `NDEBUG` definition to the resource compiler settings, nor was it set in the Visual Studio project. As a result the DLL was always build with `VS_FF_DEBUG` file flag, which was incorrect. To simplify things I chose to use `_DEBUG` in this case again, so that by default it always builds without that file flag and only if `_DEBUG` is eplicitly set it sets `VS_FF_DEBUG`. With CMake this is not set, but that's fine.

PR #124, opened 2020-12-20
Fix broken layout of virtual function tables when building with MinGW

rafalh:
Virtual function tables were not matching d3d8 because destructor pointer was inserted at offset 0. I wonder how this can work in MSVC...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,17 +3,17 @@ cmake_minimum_required(VERSION 3.12)
 project(d3d8to9 CXX)
 
 set(SRCS
-  source/d3d8to9_base.cpp
+  source/d3d8.hpp
   source/d3d8to9.cpp
-  source/d3d8to9_device.cpp
   source/d3d8to9.hpp
+  source/d3d8to9_base.cpp
+  source/d3d8to9_device.cpp
   source/d3d8to9_index_buffer.cpp
   source/d3d8to9_surface.cpp
   source/d3d8to9_swap_chain.cpp
   source/d3d8to9_texture.cpp
   source/d3d8to9_vertex_buffer.cpp
   source/d3d8to9_volume.cpp
-  source/d3d8.hpp
   source/d3d8types.cpp
   source/d3d8types.hpp
   source/d3dx9.hpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,7 @@ set(SRCS
   source/d3d8to9_texture.cpp
   source/d3d8to9_vertex_buffer.cpp
   source/d3d8to9_volume.cpp
+  source/d3d8.hpp
   source/d3d8types.cpp
   source/d3d8types.hpp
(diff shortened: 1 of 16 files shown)


crosire:
Bit dangerous since it now won't call destructors during address lookup table deletion anymore (which happens when the device is destroyed). Likely not a problem right now, since all those classes have trivial data layouts (only pointers) and therefore no destructor that does anything, but should note that with a comment then.

CookiePLMonster:
Multiple inheritance is in play here, and therefore you should end up with more than one vtable per object. I don't know if C++ guarantees the order of vtables in the object, but if it does not then it would be better to fix it by changing `QueryInterface` implementations to always typecast the returned entry to `IUnknown*`. This way you'll ensure that a proper pointer is returned.

That said, I'd check in C++ spec first if it's even possible for the compiler to reorder parent objects. If it's not possible, then this should be a non-issue in the first place.

rafalh:
I worked on this issue a bit more and I agree with @CookiePLMonster that I can be done better by making some type casts when returning  and accepting D3D8 interface pointers. Sadly it's not as easy as casting to `IUnknown*` everywhere. I don't think C++ spec says anything about vtbl order, it is probably implementation specific. I tried adding type casts to `IUnknown*` but my D3D8 app was still crashing. I was testing interface `Direct3DTexture8` and functions `GetPriority` and `GetLOD` were working correctly for me (they are declared in the base class) but when I tried to call `LockRect` I ended up in `GetLevelDesc`. Everything changed when I removed `-static` linker flag that I used to static link GCC runtime lib. After that `GetPriority` and `GetLOD` didn't work anymore. I am not sure why this flag affects vtbl.
Then I tried defining `IDirect3DTexture8` interface without multiple inheritance and added type casting pointers in `CreateTexture` to this interface. It worked fine so I think this is the way to go and make d3d8to9 compatible with GCC instead of relying on vtbl order that MSVC gives.
@crosire Are you okay with such change (adding interface classes `IDirect3D*` and inheriting from them in all implementation classes)?

CookiePLMonster:
Something doesn't add up here though, as interfaces returned from Direct3D9 must be in correct order no matter what compiler is used - else it would not be possible to use those interfaces from GCC no matter if it's d3d8to9 implementing them or if it's about stock interfaces provided by MSFT.

rafalh:
Even if they are included in Windows SDK (I believe they are not included in recent versions) they would cause a lot of conflicts with d3d9 headers that must be included in the same compilation units where we needs those interfaces. Examples would be types from d3d8/9types.h. So I think it would be difficult to reuse such headers without modifications.

crosire:
They are not, hence why d3d8types.h exists, yes. That would be the prime place to add the interface definitions too (which does sound logical to me, so please go ahead).

CookiePLMonster (approved):
Untested, but looks reasonable!

elishacloud:
I tested it and it seemed to work fine for me.

PR #130, opened 2021-03-17
Add GitHub Actions CI

Margen67:


crosire:
Not sure if this is really necessary, since the same is already implemented via AppVeyor right now.

Issue #142, opened 2022-04-18
Dear Crosire!Please help with Correcting/Remove night pixelshader effect of Operation Flashpoint 1.96


NATO2000:
Dear Crosire, have you  played Operation Flashpoint? This year is my 19 year as a player of Operation Flashpoint.

Operation Flashpoint will accompany me throughout my life.

Operation Flashpoint has a unique problem, that is, starting from the DX9 graphics card, the night image effect in the game will become black and white.
![FLASHPOINT1 2021-10-12 15-44-02-48](https://user-images.githubusercontent.com/47445654/163795634-7813af20-3aaa-4a0f-9451-b7d55a6e0c6d.jpg)
![FLASHPOINT1 2021-10-12 15-45-34-16](https://user-images.githubusercontent.com/47445654/163795673-1009b585-072f-4bd7-9483-3af5cab9af43.jpg)

[as you see,the second in game night time picture was amost black and grey]

The rendering on the DX8 graphics card is all normal.

I have checked the official technical information and player forums. They mentioned that the night rendering of the game performed abnormally on the DX9 graphics card, such as 6800GT, 7950GT and later graphics cards, but everything is normal on the FX5200 graphics card, because FX5200 is a DX8 graphics card, FX5200 The DX9 function is not complete.

What's more peculiar is that, among all the computer games in the world, Operation Flashpoint is the only game that has abnormal rendering problems with it's night images due to the upgrade of graphics card technology.

There is one ideas to solve this problem

There were two American (seem to be) players who have developed D3d8.dll for Operation Flashpoint, and matched it with Configurator.exe.

In Configurator.exe, there is such a setting:

Remove night pixelshader effect, after turning ON（not OFF） this option, the night image in the game can be rendered normally.
![QQ截图20211012154826](https://user-images.githubusercontent.com/47445654/163795710-347d7865-832c-4ce0-a5ad-9374766e3c57.jpg)

But this D3d8.dll does not make any optimizations to the frame rate and performance.
![QQ截图20211012155021](https://user-images.githubusercontent.com/47445654/163795784-3d88db7b-b193-4035-912f-1d4234162206.jpg)

Those two authors had lost contact , their E-mailboxes wasn't existing anymore

--------------------------

and then your D3D8.DLL，https://github.com/crosire/d3d8to9/releases
 provides a solution to improve the performance of the game.

This D3D8.DLL is really useful for improving the frame rate, but it does not support repairing the abnormal rendering of the game night image.

[...]

PR #143, opened 2022-04-27
Add an additional disclaimer to README regarding DirectX 8 differences on the new Windows versions, GPU drivers, etc.

HerMajestyDrMona:
As discussed in https://github.com/crosire/d3d8to9/issues/142#issuecomment-1109202254
Windows 10 / 11 is ignoring some of the game requests. For example, it results in disabled V-Sync in the DirectX 8 mode, while d3d8to9 actually follows the game requests. It may result in the false first-impression of performance with d3d8to9. I added some information about these differences to the README file, and also mentioned the amazing dxwrapper which could be useful for players who are trying to tweak older games, not only convert the API calls.
Some English tweaks or shortening would be welcome. I apologize if mentioning the other project is not allowed, I just thought it's a good solution to prevent creating many game-specific issues in this project, which are usually already resolved in dxwrapper.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ d3d8to9
 
 This is a pseudo-driver module that intends to improve compatibility and stability in games using Direct3D 8 for rendering by converting all API calls and low-level shaders to equivalent Direct3D 9 ones. By that it also opens those games to new possibilities from proven modding tools written for Direct3D 9, including [ReShade](http://reshade.me).
 
+Note that d3d8to9 does exact translation from Direct3D 8 to Direct3D 9 only. But on recent versions of Windows, depending on system settings, GPU drivers and other factors, rendering behavior in Direct3D 9 using d3d8to9 may differ from native Direct3D 8. One common issue is VSync getting enabled with d3d8to9, even though it appears inactive in native Direct3D 8, despite the application enabling it. This may give an inaccurate impression of lower performance with d3d8to9, while it actually follows what the application requests. If you want to tweak this behavior, such as forcing VSync to be off, it is recommended to use a separate tool, such as [dxwrapper](https://github.com/elishacloud/dxwrapper), which uses d3d8to9 internally for API call conversion, but adds additional configuration opportunities.
+
 ## Building
 
 You'll need Visual Studio 2013 or higher to build d3d8to9. It is recommended to install the old standalone DirectX end-user runtime, which is required for the D3DX libraries used for disassembling and assembling the shaders.
@@ -24,10 +26,6 @@ A quick overview of what some of the source code files contain:
 Any contributions to the project are welcomed, it's recommended to use GitHub [pull requests](https://help.github.com/articles/using-pull-requests/).
 
 A big shout-out to all the existing [contributors](https://github.com/crosire/d3d8to9/graphs/contributors) who worked on improving compatibility, especially [elishacloud](https://github.com/elishacloud)!
-
-## Disclaimer
-
-This module does direct translation from Driect3D8 to Direct3D9. On modern versions of Windows, depending on system settings, GPU drivers and other factors - the rendering behavior after using d3d8to9 may differ from the one in DirectX 8 mode. One of the most common issues is V-sync enabled with d3d8to9, while inactive in DirectX 8 mode. It may give a false impression of lower performance (FPS) with d3d8to9, while d3d8to9 actually follows the game requests and does not apply undocumented tweaks from the Operating System or graphics driver. If you would like to apply custom tweaks to d3d8to9, such as forcing disabling V-Sync, or enabling Anti-aliasing - it is recommended to use [dxwrapper](https://github.com/elishacloud/dxwrapper) which also uses d3d8to9 for API calls conversion, but additionally allows manual configuration.
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -25,6 +25,10 @@ Any contributions to the project are welcomed, it's recommended to use GitHub [p
 
 A big shout-out to all the existing [contributors](https://github.com/crosire/d3d8to9/graphs/contributors) who worked on improving compatibility, especially [elishacloud](https://github.com/elishacloud)!
 
+## Disclaimer
+
+This module does direct translation from Driect3D8 to Direct3D9. On modern versions of Windows, depending on system settings, GPU drivers and other factors - the rendering behavior after using d3d8to9 may differ from the one in DirectX 8 mode. One of the most common issues is V-sync enabled with d3d8to9, while inactive in DirectX 8 mode. It may give a false impression of lower performance (FPS) with d3d8to9, while d3d8to9 actually follows the game requests and does not apply undocumented tweaks from the Operating System or graphics driver. If you would like to apply custom tweaks to d3d8to9, such as forcing disabling V-Sync, or enabling Anti-aliasing - it is recommended to use [dxwrapper](https://github.com/elishacloud/dxwrapper) which also uses d3d8to9 for API calls conversion, but additionally allows manual configuration.
+
 ## License
 
 All source code in this repository is licensed under a [BSD 2-clause license](LICENSE.md).


mirh:
You shouldn't *really* blame Windows before having tested different versions in a VM with the same driver. 
Anyway W10 did indeed [introduce](https://www.vogons.org/viewtopic.php?f=8&t=47772) breaking [changes](https://github.com/elishacloud/dxwrapper/issues/134#issuecomment-1239832961) but the problem is likely a very specific one. 

It's not like that even if you put the game in fullscreen mode, d3d8 doesn't v-sync anymore. 
It is that d3d8 no longer has a real fullscreen mode to begin with (you can verify this in PresentMon). 
And be it either the DWM not allowing what would amount to double syncronization, or dx never being meant to sync non-exclusively (I would have bet on the former, but it turns out there [always](https://gamedev.net/forums/topic/468871-does-vsync-work-in-windowed-mode-d3d/468871/) was [debate](https://stackoverflow.com/questions/3915416/how-to-enable-vsync-in-d3d-windowed-app)) that cannot become your frame limiter then. 

On top of that, yes there can be other meddling. 
Windowed G-SYNC being the most famous one: when a non-fullscreen window is in-focus it will be synced to the monitor refresh rate (of note that games can still change this, even if Windows doesn't give them total control of the screen). 
Somewhat similarly IIRC (even though this really isn't going to apply to an old game) the `Composed: Flip` presentation mode should "force sync" you to integer multiples of the display frequency. 
Everything has a name and a surname eventually if you dig enough.

HerMajestyDrMona:
The goal of this description change was to reduce the number of reported issues about the "FPS lock" in d3d8to9. Although it's too early to judge, it seems like it actually helped with it.
 
Since Windows 10 / 11 actually doesn't present the d3d8 parameters in the form _expected_  by d3d8 game developers, I think it's safe to say that it's in simple words ignoring them. It doesn't really matter if the _breaking_ changes were documented or not, the final effect is the same. It's nothing really _blaming_ in it, just stating a fact that is easily understandable by all users.

I agree that it's good to know the root cause of the ~~problems~~ differences, but at the end it doesn't change much, at least in this case.

Do you propose some changes to the description in order to mention that it's caused by the lack of the _real fullscreen mode_? Or these are just extra information that the initial pull request is missing?

mirh:
> The goal of this description change was to reduce the number of reported issues about the "FPS lock" in d3d8to9.

If an application asks for vsync, of course that request should be abided. 
I get that people don't care about who's whos, and I know not all games offer a v-sync option, but (assuming I'm right of course) having a disclaimer to explain how vsync even works is really lame. 

> Since Windows 10 / 11 actually doesn't present the d3d8 parameters in the form expected by d3d8 game developers

I don't know. 
***If*** this is Windows\DWM covertly *suppressing* v-sync calls on every non-exclusive fullscreen window then that would be correct. 
But if this had always been the default/vanilla d3d behavior for windowed mode (even though not all games may have officially supported it, but that's on them to be stupid) then it is all one big misapprehension. 

> I agree that it's good to know the root cause of the problems differences, but at the end it doesn't change much, at least in this case.

It kinda does, because in that case recommending an older dll (or perhaps filling a feature request against dxwrapper), or just telling people to disable v-sync, would be the proper course of action. 

> Do you propose some changes to the description in order to mention that it's caused by the lack of the real fullscreen mode?

I'd recommend testing, because the information inferred with lots of charity from bug reports wasn't exactly clear and explicit :)

HerMajestyDrMona:
I tested 2 or 3 games on VMs with older versions of Windows (XP and 7), and I could not get more than 60 FPS on them. Empire Earth 2 for example has no switchable V-Sync option by default. This makes me assume that some games wanted V-Sync to be enabled by default. It didn't really matter back then, because hardware limitations made them run at much lower framerate.

In Windows 10 (I don't know if it started in some version or was this way since the beginning), DirectX 8 games (at least the ones I tested and the others that were reported by GitHub issues) run with more than 60 FPS by default (often 200+ since they're not very demanding). Enabling or disabling fullscreen optimizations had no effect on them.

Since d3d8to9 does an exact translation of d3d8 calls to d3d9, and makes games run with the limited FPS (V-Sync behavior), then it means that DX8 rendering behavior differs on modern version of Windows. This is probably because DX8 and older games are not really using the native API. They get converted using a similar thing to D3D9On12. I think NVIDIA indirectly [mentions it](https://nvidia.custhelp.com/app/answers/detail/a_id/3733/~/games-based-on-directx-8-and-older-versions-of-directx-will-only-run-on) on their support page.

> It kinda does, because in that case recommending an older dll (or perhaps filling a feature request against dxwrapper), or just telling people to disable v-sync, would be the proper course of action.

I'm not sure if I understood it. Which older DLL? Was there some d3d8to9 version with the V-Sync forcefully disabled? Why recommending to disable the V-Sync, if d3d8to9 does the exact and correct translation? Such recommendations don't really fit the description page when it's d3d8to9 that is doing things correctly. 

To be honest I think the current disclaimer is just fine. It is neutral, doesn't _blame_ anything and is actually accurate to what _everyone_ already knows about possible factors.

mirh:
> I tested 2 or 3 games on VMs with older versions of Windows (XP and 7) in DirectX 8 mode, and I could not get more than 60 FPS on them. 

In windowed mode? Because that's kinda the big doubt. 
I think DxWnd, 3D-Analyze and D3DWindower should be able to force it even if ALT+ENTER doesn't work (otherwise you could try games that don't require external tools like Mafia, Syberia, UT2003 or BF2142). 

> This makes me assume that some games wanted V-Sync to be enabled by default. 

Probably more like they never thought/cared for it (even the super expert devs working on GNOME had to be told why yes, it's not so easy and automatic for it to make sense)

> It didn't really matter back then, because hardware limitations made them run at much lower framerate.

Even more if your hardware risks not to keep up with VBI, vsync isn't so desirable. 
It almost seems like you think v-sync is supposed to be about frame limiting...

> In Windows 10 (I don't know if it started in some version or was this way since the beginning), 

Avoiding real fullscreen is a trend that might have started in W8 already [actually](http://web.archive.org/web/20170418171908/http://www.blitzbasic.com/Community/posts.php?topic=99477#1281001) (the source just mentions ddraw tbf, but d3d8 might have been on the table too)

> Enabling or disabling fullscreen optimizations had no effect on them.

If you read literally my first link you see that `DXPrimaryEmulation -DisableMaxWindowedMode` was the fix (before 1809).
FSO [would](https://www.pcgamingwiki.com/wiki/Windows#Fullscreen_optimizations) only apply to d3d9+, and whatever the result (`Hardware: Independent Flip` AFAICT) that shouldn't affect vsync. 

> They get converted using a similar thing to D3D9On12. 

D3D9On12 is a legit UMD/DDI ~~replacement~~ [driver](https://reshade.me/forum/general-discussion/8642-d3d9on12-is-recognized-as-dx12-but-shaders-is-still-limited-as-dx9#45147) (i.e. the driver that processes the 3d commands), it doesn't *wrap* the runtime.
As far as the other applications know (also including the windowing system and whatnot) everything is still perfectly native. 

> I think NVIDIA indirectly [mentions it](https://nvidia.custhelp.com/app/answers/detail/a_id/3733/~/games-based-on-directx-8-and-older-versions-of-directx-will-only-run-on) on their support page.

That is just a consequence of microsoft never having implemented a "dispatching api" [like](https://www.vogons.org/viewtopic.php?f=8&t=64298&start=20) for d3d9 and newer. 
And be it for technical or just economical worthiness reasons (after all, in 2015 even atom graphics should have been able to compete with 2003 gpus) nvidia didn't feel like furthering their own hybrid hackjob.

HerMajestyDrMona:
> In windowed mode? Because that's kinda the big doubt.
> I think DxWnd, 3D-Analyze and D3DWindower should be able to force it even if ALT+ENTER doesn't work (otherwise you could try games that don't require external tools like Mafia, Syberia, UT2003 or BF2142).

I'm referring to the full screen mode. I don't remember the windowed one, since I wasn't even interested in testing it back then.

> D3D9On12 is a legit UMD/DDI replacement [driver](https://reshade.me/forum/general-discussion/8642-d3d9on12-is-recognized-as-dx12-but-shaders-is-still-limited-as-dx9#45147) (i.e. the driver that processes the 3d commands), it doesn't wrap the runtime.
> As far as the other applications know (also including the windowing system and whatnot) everything is still perfectly native.

D3D9On12 uses the native d3d9 calls, but obviously does a lot of "conversions" in order to render it using DirectX 12. It results in many issues, such as [this one](https://github.com/microsoft/D3D9On12/issues/57). It's especially problematic, because new Intel drivers for 12th Gen GPUs don't support DX9 natively at all, so D3D9On12 is forced. I assume that a similar, if not the same thing was done to DirectX 8 calls since the first version of Windows 10. This is why there are some little, thankfully mostly unnoticeable differences. It's just my theory. I don't have sources to confirm it as I'm personally very little interested in DirectX. I was _forced by the nature_ to focus on it for the unofficial patch for my favorite game, but nothing else.

> You really didn't check my links... did you?
> If you restore an older version of d3d8.dll (one that didn't have exclusive fullscreen removed completely) presumably that should address your concerns.

Sorry, I thought you were referring to d3d8.dll that comes with d3d8to9 and is somewhere in older GitHub releases. Using .dlls from older versions of Windows is usually not recommended, even though this one would probably work fine. I got a bit confused by it. You mentioned that this method could be recommended in d3d8to9 description? It didn't make much sense for me to promote workarounds for unrelated things, that additionally use the same name as the project output DLL. This could confuse usually inexperienced users a lot. I also don't know if dxwrapper could legally include this old .dll that comes with older versions of Windows. From what I remember redistribution of Windows DLLs is strictly forbidden. And the final problem is that this would actually guide to use the native DirectX 8 instead of d3d8to9 (?), which in 99.9% of situations is a bad idea since d3d8to9 fixes many compatibility issues and allows old games to be actually supported by the updated graphics drivers. So promoting a method not to use the project on it's own page would be a bit strange.

> I mean, with the same scrupulous token then no disclaimer would be needed at all, isn't it?

Since d3d8to9 does things correctly, the disclaimer is not needed. However, based on reported issues about the worse performance, and questions about disabling the V-Sync, I think it's a good thing to have. Especially since it gives the opportunity to mention dxwrapper which allows ordinary users to change things they want without d3d8to9 code modification and compilation. 

I understand your goal in posting a feedback to this. It's actually a good debate, but moves to a philosophical subject instead of the practical one. For ordinary users, it doesn't matter if new Windows versions use the _fake_ fullscreen mode or not. This _fake_ fullscreen mode is still named "fullscreen" in Windows. So for the end-user it just gives a bad impression that d3d8to9 made their game run slower, while it actually makes the game behave more like it would on the old version of Windows, which is the correct and expected way.

I apologize for using some potentially inaccurate words. It's not always easy to write thoughts on the "paper", especially with a non-native vocabulary.

mirh:
> I don't remember the windowed one, since I wasn't even interested in testing it back then.

My point was that it's actually very important to know, if not any for the wording of the note. 
And in fact I'm very dumb, because in the [third page](https://www.vogons.org/viewtopic.php?f=8&t=47772&start=40) they indeed confirm my suspects (even though I'm still unsure if it's just automatically the fact of using windowed mode to forego v-sync, or rather the DWM). 

> It results in many issues, such as https://github.com/microsoft/D3D9On12/issues/57.

I know ( ͡° ͜ʖ ͡°)
But that's still exactly what I told you (and fundamentally different from what you are thinking).. it's just a "mundane bug" in the 3d processing not really dissimilar from any that could happen with a "true native" driver. It's not really that I enjoy playing the terminology-nazi cop, but having banged my head for over a decade on fixing games "imprecision snowballing into FUD" is a scourge. 

Back in topic, like.. you wouldn't blame "direct3d" if nvidia [borked](https://www.vogons.org/viewtopic.php?f=8&t=41443) their old shadow projection matrix in their drivers. 
Conversely, you would blame d3d9 instead, if starting with Vista clip planes were [busted](https://www.vogons.org/viewtopic.php?f=8&t=40913&start=20#p406003) all across the board. 

> because new Intel drivers for 12th Gen GPUs don't support DX9 natively at all

As Squall noted at the end of the thread, they do now (alternating with DXVK depending on the game)... Did you forget?
Putting even aside that it's kind of sophistry to try to distinguish between "real hardware" DDIs and "virtual" ones (as if outputting "gpu bytecode" as opposed to "d3d12 calls" was different). 

> I assume that a similar, if not the same thing was done to DirectX 8 calls since the first version of Windows 10.

Come on man. You can't really sell (literally) `d3d8_do(stuff){d3d9_do(stuff)}` repeated for every single function, as the same of `d3d8_do(stuff){reinvent_the_wheel_in_a_totally_different_api(stuff)}`. 
For as much as we know (well, at least originally) it may even be they only uniquely shimmed the fullscreen behaviour. 

> I got a bit confused by it. You mentioned that this method could be recommended in d3d8to9 description?

Yes and no. 
Assuming such and such assumptions, "new d3d8.dll" (which here is a dll name that I always ever used for the official one Windows ships) is the cause of the differences. And in mentioning this fact in the note, "the method of using an older dll existing" would be automatically more or less implied. 

Not that I really know a reason people shouldn't use d3d8to9 anyway, but still. 

>  I also don't know if dxwrapper could legally include this old .dll that comes with older versions of Windows.

No, I meant it as in: dxwrapper takes care of restoring exclusive fullscreen behaviour into the new d3d8.dll (@elishacloud). 

> Since d3d8to9 does things correctly, the disclaimer is not needed.

Yesn't, this is kinda of an "ethical" dilemma. 
It all depends on whether the "contract" with the system you believe is being "signed", is about technically "v-sync itself" or just the practical promise that "tearing won't happen". 

> So for the end-user it just gives a bad impression that d3d8to9 made their game run slower, while it actually makes the game behave more like it would on the old version of Windows

I mean, did I already said that v-sync is pretty crappy and it shouldn't be relied on for "gaming frametimes" purposes? 
Going from the DWM-friendly `DXMaximizedWindowedMode` to good old legacy exclusive bitblt v-sync (despite being an upgrade in many compatibility aspects) could also still be argued to be a downgrade. Even though it's not further impossible to promote it to the flip model presentation and whatnot.

HerMajestyDrMona:
> As Squall noted at the end of the thread, they do now (alternating with DXVK depending on the game)... Did you forget?

No, their newest drivers simply don't fix the issue. I asked a few 12gen users to test it, and it didn't help in any case. I don't have a possibility to test it myself, and I will probably avoid Intel for the rest of my life. So I don't know if this "restored native support" is real or not.

> For as much as we know (well, at least originally) it may even be they only uniquely shimmed the fullscreen behaviour.

Possible. They didn't allow users to "disable fullscreen optimizations" on Win 8.1 (or maybe even first versions of 10). This option came up later, so maybe they started cutting / disabling things considering they will never be used again.

It's just pity that the "disable fullscreen optimizations" simply don't restore the native DX8 fullscreen behavior now. As noted earlier, it was probably not worth to support it. Especially when in most cases it just doesn't matter.

mirh:
> No, their newest drivers simply don't fix the issue.

I mean, it could even be that their replacement driver has in its turn that bug. For as much as odd. 

> So I don't know if this "restored native support" is real or not.

As I said, you should stop to focus on what kind of "actual backend" a legitimate driver has (in fact, you may even argue supporting D3D9On12 is a super good thing, if not any just because it's at least open source). 
Though you have a point, that if the claim is "the original driver or DXVK has been restored" and then you still see the bug (even though neither of them assumedly experienced it) one of the premises must be wrong eventually. 

...
Anyway long story short the factors should be (underlining again that version delimitation should be cross-checked):
* every d3d8 windowed application **always** [had](https://web.archive.org/web/20030213002025/http://msdn.microsoft.com:80/library/en-us/dx8_c/directx_cpp/Graphics/Reference/CPP/D3D/Structures/d3dpresent_parameters.asp) `D3DPRESENT_INTERVAL_DEFAULT(0)` (and windowed fullscreen [isn't](https://forum.thegamecreators.com/thread/34795) new?)
* fast forward to our days W10 rtm [added](https://www.vogons.org/viewtopic.php?f=8&t=47772) or turned on by default `Direct3D8EnableMaximizedWindowedModeShim`
* 1607 fixed some [supposed](https://github.com/ThirteenAG/WidescreenFixesPack/issues/125#issuecomment-232681468) problem (with.. borders?)
* 1809 dropped any support for exclusive modes

Squall-Leonhart:
The D3D8 vsync issue here is not a driver issue, its a runtime issue within the Windows 10/11 D3D8 dll

mirh:
Yes, I had been trying to underline above the difference between the actual device driver, and the runtime (I should probably have referenced [this](https://learn.microsoft.com/en-us/windows-hardware/drivers/display/windows-vista-and-later-display-driver-model-architecture) though, rather than crosire shallowly using the former term for everything)... 
Anyway, this is nothing else than what @UCyborg reported already 4 years ago in the first link I posted.

PR #154, opened 2023-05-09
Fix ZBIAS value

elishacloud:
`D3DRS_ZBIAS` in Direct3D8 is converted to `D3DRS_DEPTHBIAS` in Direct3D9 because `D3DRS_ZBIAS` is no longer available.  However, there was an issue with the way it was converted.

In Direct3D8 "polygons with a high z-bias value appear in front of polygons with a low value, without requiring sorting for drawing order." See [here](https://learn.microsoft.com/en-us/previous-versions/ms886344(v=msdn.10)).

In Direct3D9 "a higher z-bias value increases the likelihood that the polygons you render will be visible when displayed with other coplanar polygons."  See [here](https://learn.microsoft.com/en-us/windows/win32/direct3d9/depth-bias).

So in both cases a higher bias will help to draw polygons in front of polygons with a lower bias.

However, ZBIAS was multiplied by a negative number.  This means what was a larger bias in Direct3D8 becomes a smaller bias after converting it to Direct3D9.  This is certainly not what is expected.  Also, this resulted in a negative number, which caused issues in some games.  Silent Hill 2, for example, on Nvidia cards will not display the 2D fog because of the incorrect bias conversion.

Furthermore, the `Set` value was multiplied by 0.000005f and the `Get` value was multiplied by 500000.0f.  However, the `Get` value should be multiplied by 200000.0f to bring the original value back.  This has been fixed in this PR.

_BTW: I also removed an unnecessary call to `memcpy()`._


universal963:
Hi there. Unfortunately this fix somehow breaks Command & Conquer Renegade on my machine. It causes the bullet hole on the ground not shown and some part of ground not looking right, while the real D3D8 (Windows 10 22H2) and d3d8to9 commit 1753ae1 (one commit before this) behave right.
Here are some screenshots:
real d3d8:
d3d8
d3d8to9 commit 1753ae1:
d3d8to9
d3d8to9 latest commit:
d3d8to9_latest

elishacloud:
@universal963, can you try making the change listed here:  https://github.com/elishacloud/d3d8to9/commit/9066eb9574d06a5cde4a4ddacfa8ac672aa2c9f3

universal963:
@elishacloud It seems to be right with that change.
d3d8to9_response1

PR #156, opened 2023-05-11
Revert part of the ZBIAS value fix

elishacloud:
It seems like ZBIAS does not work as expected by the documentation.  Reverting part of the change until I have time to do more testing and come up with a better fix.

See comments [here](https://github.com/crosire/d3d8to9/pull/154#issuecomment-1544266806).


elishacloud:
It seems like the ZBIAS is not well supported.  See [ Depth Bias not supported?](https://gamedev.net/forums/topic/250871-depth-bias-not-supported/2507876/) and [Why ZBIAS is not a good thing](http://tomsdxfaq.blogspot.com/2002_07_01_tomsdxfaq_archive.html#79344425).

mirh:
Also discussed [here](https://www.vogons.org/viewtopic.php?t=34931&start=1920) @ucyborg

Issue #155, opened 2023-05-10
D3D8.1 Pixel and Vertex Shader Lifetimes Seem To Be Incorrect


nv-ajaus:
I've noticed that D3D9 Devices and D3D9 Pixel/Vertex Shaders have been leaking with some applications. Looking through the D3D8to9 code, I think there is some mismanagement of the D3D9 Shader refcounts because of some behavioral differences between D3D8.1 and D3D9. 

In D3D8.1 Pixel and Vertex Shaders do not increase the refcount of the device they are created from, but they do in D3D9 (This was confirmed through testing on both ends). Similarly, in D3D9, a device reset preserves the Shader objects, but the Microsoft documentation for D3D9 Lost Devices mentions that "In Direct3D 9, vertex shaders and pixel shaders don't need to be re-created after reset. They will be remembered. In previous versions of DirectX, a lost device required shaders to be re-created." (https://learn.microsoft.com/en-us/windows/win32/direct3d9/lost-devices). 

D3D8to9 only calls for a release on D3D9 Shader objects when there is a call to DeletePixelShader or DeleteVertexShader, but I think to maintain compatible or expected behavior, the device object in D3D8to9 should keep track of the handles of any shaders created through it and explicitly release them on a release or reset call to the device.

nv-ajaus:
Closed with PR #158

PR #158, opened 2023-06-10
Shader lifetime improvement

nv-ajaus:
This is a proposal to fix the issue outlined in https://github.com/crosire/d3d8to9/issues/155. Basically, D3D8.1 and D3D9 lifetimes have different rules but d3d8to9 currently doesn't follow the D3D8.1 behavior (which is that shaders share a lifetime with the device that created them and are implicitly destroyed). This causes some D3D8 applications to leak shader and device objects even if they are well behaved under plain D3D8.x.

Here is a test that shows this behavior:
[d3d8_refcounttest.zip](https://github.com/crosire/d3d8to9/files/11711248/d3d8_refcounttest.zip)

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -1526,7 +1526,8 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::DeleteVertexShader(DWORD Handle)
 	const DWORD HandleMagic = Handle << 1;
 	VertexShaderInfo *const ShaderInfo = reinterpret_cast<VertexShaderInfo *>(HandleMagic);
 
-	if (ShaderInfo->Shader != nullptr) {
+	if (ShaderInfo->Shader != nullptr) 
+	{
 		ShaderInfo->Shader->Release();
 		VertexShaderHandles.erase(reinterpret_cast<DWORD>(ShaderInfo->Shader));
 	}
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -49,13 +49,20 @@ ULONG STDMETHODCALLTYPE Direct3DDevice8::AddRef()
 {
 	return ProxyInterface->AddRef();
 }
+
 ULONG STDMETHODCALLTYPE Direct3DDevice8::Release()
 {
+	ULONG LastRefCount = ProxyInterface->Release();
+
 	// Shaders are destroyed alongside the device that created them in D3D8 but not in D3D9
 	// so we Release all the shaders when the device releases to mirror that behaviour
-	ReleaseShaders();
-
-	ULONG LastRefCount = ProxyInterface->Release();
+	if (LastRefCount !=0 && LastRefCount == (VertexShaderAndDeclarationCount + PixelShaderHandles.size()))
+	{
+		ProxyInterface->AddRef();
+		ReleaseShaders();
+		LastRefCount = ProxyInterface->Release();
+		assert(LastRefCount == 0);
+	}
 
 	if (LastRefCount == 0)
 		delete this;
@@ -1427,6 +1434,7 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::CreateVertexShader(const DWORD *pDecl
 		ShaderInfo = new VertexShaderInfo();
 
 		hr = ProxyInterface->CreateVertexShader(static_cast<const DWORD *>(Assembly->GetBufferPointer()), &ShaderInfo->Shader);
+		VertexShaderAndDeclarationCount++;
 
 		Assembly->Release();
 	}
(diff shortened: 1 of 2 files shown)


source/d3d8to9_device.cpp
```diff
@@ -51,6 +51,10 @@ ULONG STDMETHODCALLTYPE Direct3DDevice8::AddRef()
 }
 ULONG STDMETHODCALLTYPE Direct3DDevice8::Release()
 {
+	// Shaders are destroyed alongside the device that created them in D3D8 but not in D3D9
+	// so we Release all the shaders when the device releases to mirror that behaviour
+	ReleaseShaders();
```
crosire, line 56: I'm a bit unhappy with this, as it means shaders are destroyed even when an app does something like this, which should technically be fine:
```c++
device->CreateVertexShader();
device->AddRef();
device->Release();
device->SetVertexShader(); // Oops, shader handle is broken now
```
But since the D3D9 shader objects will hold a device reference, the `LastRefCount == 0` condition would never be triggered if an attempt was made to instead release shaders in there.
A potential solution (albeit very hacky) might be this:
```c++
ULONG STDMETHODCALLTYPE Direct3DDevice8::Release()
{
    ULONG LastRefCount = ProxyInterface->Release();

    if (LastRefCount == (VertexShaderHandles.size() + PixelShaderHandles.size())
    {
        ProxyInterface->AddRef();
        ReleaseShaders();
        LastRefCount = ProxyInterface->Release();
        assert(LastRefCount == 0);
    }

    if (LastRefCount == 0)
        delete this;

    return LastRefCount;
}
```

source/d3d8to9_device.cpp
```diff
@@ -2134,3 +2152,21 @@ void Direct3DDevice8::ApplyClipPlanes()
...
+		reinterpret_cast<T>(Handle)->Release();
+	}
+	ShaderHandles.clear();
+}
+
+void Direct3DDevice8::ReleaseShaders()
+{
+	ReleaseShaderSet<IDirect3DPixelShader9*>(PixelShaderHandles);
+	SetPixelShader(0);
+	ReleaseShaderSet<IDirect3DVertexShader9*>(VertexShaderHandles);
```
crosire, line 2170: Doesn't this leak the vertex declaration objects now? I think `VertexShaderHandles` needs to track the actual handles generated in `CreateVertexShader` and then this can simply call `DestroyVertexShader` on them (same thing for pixel shader handles). That will also do the `SetVertexShader(0)` calls internally already and thus avoid some code duplication.

crosire (commented):
Thank you for addressing this!

source/d3d8to9_device.cpp
crosire: If tracking the vertex declaration as per below, that condition would change to `if (LastRefCount == (VertexShaderHandles.size() * 2 + PixelShaderHandles.size())`.

nv-ajaus: Good point, I hadn't thought of that use case. Unfortunately I think the solution is even hackier, since not all D3D9 CreateVertexDeclaration calls are matched to a D3D9 CreateVertexShader call (because no function pointer can legally be passed to use the fixed function pipeline), the reference count delta for each call to D3D8 CreateVertexShader can be 1 or 2. I added some manual reference counting for Vertex Shaders and Declarations on the D3D8to9 end to accommodate this.

source/d3d8to9_device.cpp
nv-ajaus: It does, thank you for noticing that! The host application I was using was manually deleting vertex shaders/declarations but not pixel shaders so I never noticed it in testing. I made the change you recommended.

source/d3d8to9_device.cpp
```diff
@@ -2134,3 +2173,18 @@ void Direct3DDevice8::ApplyClipPlanes()
...
+		DeletePixelShader(Handle);
+	}
+	PixelShaderHandles.clear();
+	for (auto Handle : VertexShaderHandles)
+	{
+		DeleteVertexShader(Handle);
+	}
+	VertexShaderHandles.clear();
+	VertexShaderAndDeclarationCount = 0;
+}
```
elishacloud, line 2190: This does not reset all the values for the shaders.  d3d8to9 also caches shaders so we should reset the cache like this:

```c++
CurrentVertexShaderHandle = 0;
CurrentPixelShaderHandle = 0;
```

Also, I am not sure if Direct3D9 remembers which shader was set before the Release and if we should reset that also?

```c++
ProxyInterface->SetPixelShader(nullptr);
ProxyInterface->SetVertexShader(nullptr);
```

crosire: This is handled by the calls to `Delete(Pixel/Vertex)Shader` (which will call `Set(Pixel/Vertex)Shader(0)` for the respective current handle once encountered). So it should actually reset those values already.

PR #168, opened 2024-03-10
Fix memory leak with assembly buffer

elishacloud:
Fixes a few places where the assembly buffer is not released correctly.


CookiePLMonster:
`Microsoft::WRL:ComPtr` keeps winning with its scope-based lifetime for COM objects.

crosire:
That's why this likely slipped through, I'm too used to using smart COM pointers =D

elishacloud:
Good point.  I should have switched them over to that when I fix them.  Oh well.  They are fixed now.  If I create another PR in the future around this area maybe I can switch them over to smart pointers also.

PR #173, opened 2024-09-07
Return D3D_OK for Set/GetRenderState calls with D3DRS_ZVISIBLE

WinterSnowfall:
Tests on Windows XP have shown that these particular calls don't actually error out for D3DRS_ZVISIBLE, but simply return D3D_OK. Also, the default value of D3DRS_ZVISIBLE is 0.


elishacloud:
I believe this is a good change.

I just tested and it seems like `SetRenderState()` and `GetRenderState()` should always return DD_OK as long as the State is less than or equal to 255, even if unsupported States are specified.  In fact, it might even return DD_OK if the state is greater than 255.  I don't see anything in the documentation about it returning anything bu DD_OK.

PR #175, opened 2024-09-11
Fix resource leak and prevent pure device creation

elishacloud:
This fixes a resource leak that happens in `GetTexture()` because we query for the interface after getting the texture and never release the extra reference that creates.

Also, this prevents the game from creating a pure device.  d3d8to9 won't work with a pure device because it requires many `Get` functions that won't work if a game creates a pure device.

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -851,14 +851,17 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::GetTexture(DWORD Stage, IDirect3DBase
 		case D3DRTYPE_TEXTURE:
 			BaseTextureInterface->QueryInterface(IID_PPV_ARGS(&TextureInterface));
 			*ppTexture = ProxyAddressLookupTable->FindAddress<Direct3DTexture8>(TextureInterface);
+			TextureInterface->Release();
 			break;
 		case D3DRTYPE_VOLUMETEXTURE:
 			BaseTextureInterface->QueryInterface(IID_PPV_ARGS(&VolumeTextureInterface));
 			*ppTexture = ProxyAddressLookupTable->FindAddress<Direct3DVolumeTexture8>(VolumeTextureInterface);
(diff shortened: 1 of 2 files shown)


Polymega:
> Maybe pure devices are more common in Direct3D9 games than in Direct3D8 or maybe these are the reason for a number of games that already fail or have issues with d3d8to9 that hasn't been investigated yet. I know several.

Wonder if this would fix Silent Hill 3's unique DX8 shaders? Right now, all advanced beauty shaders made by SH3 don't render at all when using d3d8to9.

elishacloud:
> Not all `Get*` functions are actually disabled, rather only the ones affecting state, if docs are to be believed. See [here](https://learn.microsoft.com/en-us/windows/win32/dxtecharts/directx-9-frequently-asked-questions?redirectedfrom=MSDN#what-is-the-purpose-of-the-d3dcreate--puredevice-flag--).

Ok, I removed the code that prevents pure device creation.

source/d3d8to9_device.cpp
```diff
@@ -851,14 +851,17 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::GetTexture(DWORD Stage, IDirect3DBase
 		case D3DRTYPE_TEXTURE:
 			BaseTextureInterface->QueryInterface(IID_PPV_ARGS(&TextureInterface));
 			*ppTexture = ProxyAddressLookupTable->FindAddress<Direct3DTexture8>(TextureInterface);
+			TextureInterface->Release();
```
crosire, line 854: I'd maybe just call `BaseTextureInterface->Release()` in all case blocks (including the `default` actually, for completeness, not that it would ever be reached). Functionally there is no difference, but logically the code is using the queried interface for the proxy lookup (which is what is returned), so looks cleaner to keep the reference on those particular variables I think.

elishacloud:
Ok, good call.  I fixed it to release the base texture in each block.

PR #177, opened 2024-09-14
Remove CreateImageSurface R8G8B8 swap with X8R8G8B

WinterSnowfall:
I'm guessing this format swap is a remnant from the initial code, if git blame is to be believed. It should no longer be needed now, as image surfaces with R8G8B8 will be created just fine in D3DPOOL_SCRATCH, with the current fallback logic.

The swap usually works fine anyway, but has the drawback of causing minor issues at times, such as corrupting savegame screenshots in [Chrome](https://www.pcgamingwiki.com/wiki/Chrome). Keeping R8G8B8 as is solves this problem.

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -414,14 +414,6 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::CreateImageSurface(UINT Width, UINT H
 
 	*ppSurface = nullptr;
 
-	if (Format == D3DFMT_R8G8B8)
-	{
-#ifndef D3D8TO9NOLOG
-		LOG << "> Replacing format 'D3DFMT_R8G8B8' with 'D3DFMT_X8R8G8B8' ..." << std::endl;
-#endif
-		Format = D3DFMT_X8R8G8B8;
-	}
-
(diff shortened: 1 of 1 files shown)


elishacloud:
Definitely, if the game requests R8G8B8 (a 24bit surface) and instead we create it as X8R8G8B (a 32bit surface) then that will cause problems for the game if they lock it (or try and get the device context for it) because they will either be putting the wrong content into the surface or reading the wrong content from the surface.  The Lock and GetDC functions would need to be emulated for that to work.

In the mean time, I agree with not replacing the surface format.  If creating the surface as a D3DPOOL_SCRATCH works that is even better.  Otherwise, if there is a game with an issue then we should add emulation for these functions rather than just replacing the surface format.

WinterSnowfall:
> If creating the surface as a D3DPOOL_SCRATCH works that is even better. 

In the case of Chrome at least, it does. The game uses a R8G8B8 to create a vegetation lightmap.

I think it was either "Hidden & Dangerous Deluxe" or "The Fellowship of the Ring" which was also creating some R8G8B8 surfaces. Otherwise it's not really commonly used in d3d8.

PR #179, opened 2024-09-21
Implement Direct3DDevice8::GetInfo

WinterSnowfall:
Implementation of Direct3DDevice8::GetInfo, which partially maps onto Direct3DDevice8::CreateQuery followed by queries on a IDirect3DQuery9 object.

The specific details of the implementation, especially the response codes have been recreated based on the information available in the MS DX8.1 SDK. Unfortunately, the documentation you can find online on the topic is not at all extensive.

This is largely relevant on Nvidia for D3DDEVINFOID_VCACHE type queries, since some games expect *some* data there and native drivers do indeed return non-zero information ([see here for some samples](https://drive.proton.me/urls/1WHHKT8P0C#cbJ6R2bTCCF8)). Beside games, VCACHE queries have also been observed to be initiated by D3D9 D3DX libraries, though I am unsure if this is also the case in D3D8.

Needs a bit of testing and validation, submitting as draft for now.


WinterSnowfall:
Verified to be working properly at least on D3DDEVINFOID_VCACHE queries. I'm unsure if MinGW needs any additional header definitions, but MSVC is fine as is.

source/d3d8to9_device.cpp
```diff
@@ -971,15 +971,72 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::ValidateDevice(DWORD *pNumPasses)
...
 #ifndef D3D8TO9NOLOG
 	LOG << "Redirecting '" << "IDirect3DDevice8::GetInfo" << "(" << this << ", " << DevInfoID << ", " << pDevInfoStruct << ", " << DevInfoStructSize << ")' ..." << std::endl;
 #endif
 
-	return S_FALSE;
+	if (pDevInfoStruct == nullptr || DevInfoStructSize == 0)
+		return D3DERR_INVALIDCALL;
+
+	HRESULT hr;
+	IDirect3DQuery9 *pQuery;
```
elishacloud, line 982: This should be initialized to nullptr since you check below if it is null.

WinterSnowfall: Done, sorry for the oversight. I'm used to dealing with some form of smart pointers usually, so this slipped my mind.

elishacloud: NP.  Thanks.

PR #180, opened 2024-09-21
Validate depth formats on use with CopyRects

WinterSnowfall:
The D3D8 specs also say, with regards to CopyRects:

> This method cannot be applied to surfaces whose formats are classified as depth stencil formats.

I have checked this claim against native drivers and they do indeed validate it.

### DIFF
--- a/source/d3d8types.cpp
+++ b/source/d3d8types.cpp
@@ -15,6 +15,16 @@ bool SupportsPalettes()
 	return hasPalette;
 }
 
+bool IsDepthStencil(D3DFORMAT &Format) {
+	return Format == D3DFMT_D16_LOCKABLE
+		|| Format == D3DFMT_D16
+		|| Format == D3DFMT_D32
+		|| Format == D3DFMT_D15S1
+		|| Format == D3DFMT_D24X4S4
+		|| Format == D3DFMT_D24S8
+		|| Format == D3DFMT_D24X8;
+}
+
 static UINT CalcTextureSize(UINT Width, UINT Height, UINT Depth, D3DFORMAT Format)
 {
 	switch (static_cast<DWORD>(Format))
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -445,6 +445,9 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::CopyRects(IDirect3DSurface8 *pSourceS
 	if (SourceDesc.Format != DestinationDesc.Format)
 		return D3DERR_INVALIDCALL;
 
+	if (IsDepthStencil(SourceDesc.Format))
+		return D3DERR_INVALIDCALL;
+
 	HRESULT hr = D3DERR_INVALIDCALL;
 
 	if (cRects == 0)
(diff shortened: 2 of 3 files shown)


source/d3d8types.cpp
```diff
@@ -15,6 +15,16 @@ bool SupportsPalettes()
...
 
+bool IsDepthStencil(D3DFORMAT &Format) {
+	return Format == D3DFMT_D16_LOCKABLE
+		|| Format == D3DFMT_D16
+		|| Format == D3DFMT_D32
+		|| Format == D3DFMT_D15S1
+		|| Format == D3DFMT_D24X4S4
+		|| Format == D3DFMT_D24S8
+		|| Format == D3DFMT_D24X8;
+}
```
elishacloud, line 26: I am not sure if these are used by any d3d8 game.  But there are a few other depth stencil surfaces missing.  Maybe it makes sense to just add them all now:

* D3DFMT_S1D15 (72)
* D3DFMT_S8D24 (74)
* D3DFMT_D24S8 (75)
* D3DFMT_X8D24 (76)
* D3DFMT_X4S4D24 (78)
* D3DFMT_D32F_LOCKABLE (82)
* D3DFMT_D24FS8 (83)
* D3DFMT_D32_LOCKABLE (84)

WinterSnowfall: None of these (except D3DFMT_D24S8, which I've already included above) are part of the D3D8 standard or D3DFORMAT enum, so the likelihood of any D3D8 game using them is doubtful. Some of them are included and used in D3D9, however, as you are aware, CopyRects is D3D8 specific.

elishacloud: Keep in mind that some of the stencil formats were added to video cards after DX 8.1.  They may still work with video cards.  I don't think adding all known ones is bad.

However, maybe a better way, rather than using a function to check specific formats, is to just ask d3d9 if it is a stencil buffer by using the following code:

```c++
SourceDesc.Usage == D3DUSAGE_DEPTHSTENCIL
```

WinterSnowfall: > However, maybe a better way, rather than using a function to check specific formats, is to just ask d3d9 if it is a stencil buffer by using the following code:
> 
> ```c++
> SourceDesc.Usage == D3DUSAGE_DEPTHSTENCIL
> ```

That indeed sounds a lot better than adding formats which were not known by the d3d8 runtime. I will need to test this behavior against native d3d8, but will implement it if it checks out. Thank you for the suggestion.

WinterSnowfall: @elishacloud Sadly, that doesn't seem to be a correct approach. I checked, and you can create a texture and underlying surfaces that have a depth stencil format without specifying the D3DUSAGE_DEPTHSTENCIL flag. Those surfaces would then not fail the CopyRects validation, which is not the intended behavior. It seems to be format not usage specific, as the specs mention.

If you want, I can test to see if CopyRects validates any of the extra formats you've specified, however I doubt that is the case. The ones I've already included should be sufficient to prevent any unwanted CopyRects behavior in d3d8 games. If we notice exceptions to the rule, we can always refine it later.

elishacloud: > I checked, and you can create a texture and underlying surfaces that have a depth stencil format without specifying the D3DUSAGE_DEPTHSTENCIL flag.

Ok, too bad.  I guess the only way is to check the surface format.  Thanks!

Issue #159, opened 2023-06-16
Latest update crashes Silent Hill 2 in SetVertexShader()


elishacloud:
The latest check-in from PR #158 crashes Silent Hill 2 in `SetVertexShader()` after changing resolutions.

As far as I can tell Silent Hill 2 does not recreate Vertex or Pixel shaders after a Reset().  It seems that Silent Hill 2 assumes that the shaders are still available even after the Reset().

Sometimes Microsoft's documentation is wrong.  It seems like this may be the case here because Silent Hill 2 reuses shaders created before the Reset().

HerMajestyDrMona:
For me Empire Earth 2 crashes when using ALT+ENTER hotkey (switching between the windowed mode / full screen). It only happens in game (not menu). I assume the reason might be similar, even though changing screen resolution seem to work.

HerMajestyDrMona:
@elishacloud This is strange. It stops crashing for me after enabling Antialiasing (after adding some code that you sent me a few years ago). Just as if there was something added with Antialiasing stuff that fixes crashing caused by PR #158.

nv-ajaus:
Addressed this in #161 , I was wondering if this was related to D3D8 mixing FVF calls and programmable shader creation together in CreateVertexShader(), or something to do with the Vertex Declaration objects created in d3d8to9 but it seems some d3d8 applications just use shaders across a D3DERR_DEVICELOST return from Present() and Reset().

elishacloud:
Yeah, Microsoft's documentation is not always correct with these older DirectX APIs.  I run into this from time-to-time.

CookiePLMonster:
> Yeah, Microsoft's documentation is not always correct with these older DirectX APIs. I run into this from time-to-time.

FWIW MS folks who own DirectX-related pages are one of the most responsive out of all docs maintainers so if you submit a doc fix they will certainly take a look.

PR #181, opened 2024-09-25
Reference count fix for Shaders and StateBlocks to prevent a crash when exiting games

elishacloud:
This will fix PR #159 so shaders and state blocks are correctly released once the device reference count reaches 1.  The previous code would cause a crash on exit in a couple of games, including _Indiana Jones and the Emperor's Tomb_ and _[State of Emergency](https://github.com/elishacloud/dxwrapper/issues/254#issuecomment-2372713179)_.

This fix does the following things:

1. It fixes the ref count issue in both `AddRef()` and `Release()` to return the correct value based on how the native Direct3D8 works.
2. It cleans up some code related to the shader count in `CreateVertexShader()`
3. It fixes a crash in Windows 7 64bit caused by erasing an item in a map while looping through the same map.

I tested this with a couple of games alongside the native d3d8 to ensure it does not crash and it works the same as the native d3d8.

_Note: it appears that you need to get the current device reference count by call `AddRef()` and then `Release()` before releasing the shaders and state blocks.  If you call `Release()` first then it can lead to a crash in some games._

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -2223,7 +2223,7 @@ void Direct3DDevice8::ReleaseShadersAndStateBlocks()
 	VertexShaderAndDeclarationCount = 0;
 	while (!StateBlockTokens.empty())
 	{
-		DWORD Token = *VertexShaderHandles.begin();
+		DWORD Token = *StateBlockTokens.begin();
 		DeleteStateBlock(Token);
 	}
 }
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -2210,21 +2210,20 @@ void Direct3DDevice8::ApplyClipPlanes()
 
 void Direct3DDevice8::ReleaseShadersAndStateBlocks()
 {
-	// Since DeletePixelShader, [...] each erases the handle from the unordered set we need to use a temporary set
-	std::unordered_set<DWORD> tmpPixelShaderHandles = std::move(PixelShaderHandles);
-	for (auto Handle : tmpPixelShaderHandles)
+	while (!PixelShaderHandles.empty())
 	{
+		DWORD Handle = *PixelShaderHandles.begin();
 		DeletePixelShader(Handle);
 	}
-	std::unordered_set<DWORD> tmpVertexShaderHandles = std::move(VertexShaderHandles);
-	for (auto Handle : tmpVertexShaderHandles)
+	while (!VertexShaderHandles.empty())
 	{
+		DWORD Handle = *VertexShaderHandles.begin();
 		DeleteVertexShader(Handle);
 	}
-	std::unordered_set<DWORD> tmpStateBlockTokens = std::move(StateBlockTokens);
 	VertexShaderAndDeclarationCount = 0;
-	for (auto Token : tmpStateBlockTokens)
+	while (!StateBlockTokens.empty())
 	{
+		DWORD Token = *VertexShaderHandles.begin();
 		DeleteStateBlock(Token);
 	}
 }
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -10,8 +10,8 @@
 
 struct VertexShaderInfo
 {
-	IDirect3DVertexShader9 *Shader;
-	IDirect3DVertexDeclaration9 *Declaration;
+	IDirect3DVertexShader9 *Shader = nullptr;
+	IDirect3DVertexDeclaration9 *Declaration = nullptr;
 };
 
 Direct3DDevice8::Direct3DDevice8(Direct3D8 *d3d, IDirect3DDevice9 *ProxyInterface, BOOL EnableZBufferDiscarding) :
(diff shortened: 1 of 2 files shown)


elishacloud:
This PR is ready to check-in.  It fixes a crash in Windows 7 caused by modifying the unordered map in the middle of looping through the same unordered map.  It also fixes the reference count in both `AddRef()` and `Release()` to make it match what it would be on native Direct3D8.

WinterSnowfall:
>     2. It fixes a state block issue where it should not be inserting a token in `EndStateBlock()`.

I don't think this is incorrect. There are two ways of creating a state block in D3D8:
- `CreateStateBlock()`
- `BeginStateBlock()` followed by `EndStateBlock()`

You'll notice both `CreateStateBlock()` and `EndStateBlock()` take a `DWORD* pToken` which is then returned to the calling app, thus both will create valid ref counted state blocks in D3D9, with separate tokens at D3D8 level.

If you want, I can write a test to check it explicitly, but based on the above described behavior it should be pretty clear how it works.

source/d3d8to9_device.cpp
```diff
@@ -775,12 +792,7 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::EndStateBlock(DWORD *pToken)
 	if (pToken == nullptr)
 		return D3DERR_INVALIDCALL;
 
-	HRESULT hr = ProxyInterface->EndStateBlock(reinterpret_cast<IDirect3DStateBlock9 **>(pToken));
```
WinterSnowfall, line 778: Please consider keeping this in place, as per above.

WinterSnowfall (changes requested):

source/d3d8to9_device.cpp
elishacloud: Ok, good catch.  I had not realized that state blocks can be added that way.  Fixed!

elishacloud:
> There are two ways of creating a state block in D3D8.  [...] You'll notice both `CreateStateBlock()` and `EndStateBlock()` take a `DWORD* pToken` which is then returned to the calling app, thus both will create valid ref counted state blocks in D3D9, with separate tokens at D3D8 level.

Ok, good catch.  I have fixed this issue.

> P.S.: Some D3D8 games are known to leak state blocks and will crash in certain situations even with native D3D8.

The games I tested did not use StateBlocks at all.  One used only Vertex Shaders.  Plus, I literally just started the game and exited immediately on the opening menu and it crashed.  You don't even need to play the game to see the crash.

The issue with Windows 7 64bit was that when an item in map is modified while looping through the map then the next iteration though the loop would retrieve, and thus send, an invalid handle to `DeleteVertexShader()` causing a crash.  I am not completely sure why, but I believe that for some reason in Windows 7 64bit when erasing an item from the map it would move the map to a new address space, however, the loop would still continue to loop through the map as if it was still in the old address space, thus causing an invalid handle to be retrieved and sent.

Note: If the game crashes even with native d3d8 then it would not be an issue with d3d8to9.

elishacloud:
> If you want, I can write a test to check it explicitly, but based on the above described behavior it should be pretty clear how it works.

You are welcome to test this.  More testing is always good.  But I did already ran this through several games and checked the reference count before and after every call to any create function and every call to the Release and AddRef functions.  I tested both native d3d8 and I tested with d3d8to9 to verify the references match.

WinterSnowfall:
> Ok, good catch. I have fixed this issue.

Thank you.

> The games I tested did not use StateBlocks at all. One used only Vertex Shaders. Plus, I literally just started the game and exited immediately on the opening menu and it crashed. You don't even need to play the game to see the crash.

Well, that is indeed a strange behavior with respect to iterators. But, while clearing remaining shaders/state blocks, perhaps it's worth simply doing a decrementing index based for loop while retaining the vector's original size in a separate variable? Surely that's less expensive than copying the entire vector. 

Something like (an example with state blocks only):
```
	UINT stateBlockTokensEndIndex = StateBlockTokens.size() - 1;
	for (UINT i = stateBlockTokensEndIndex; i >= 0; i--)
	{
		DeleteStateBlock(StateBlockTokens[i]);
	}
	StateBlockTokens.clear();
```

Anyway, it's only a suggestion, I expect even the current code works just fine as well.

Cossacks II is the game to test for anything state block related, or perhaps Supreme Ruler 2010/2020, as the latter spuriously creates state blocks every frame. This is just FYI, if you ever need to test anything state block related.

> Note: If the game crashes even with native d3d8 then it would not be an issue with d3d8to9.

Indeed. My point was mostly that crashes due to various leaks, especially on alt+tab-ing, are not unheard of.

> You are welcome to test this. More testing is always good. But I did already run this through several games and checked the reference count before and after every call to any create function and every call to the Release and AddRef functions. I tested both native d3d8 and I tested with d3d8to9 to verify the references match.

I was only referring to the state blocks, but I'm pretty sure I have checked this myself as well in the past. Otherwise I see no problem with the shader related changes, if they improve compatibility.

elishacloud:
> Edit: Nevermind on the initial suggestion, missed that sets were used here, not vectors.

Yeah, unfortunately std::unordered_set don't support indexing.  I don't think using std::move is too bad.  Keep in mind the code only runs when the whole device is being deleted anyways.

> Otherwise I see no problem with the shader related changes, if they improve compatibility.

Thanks for checking.  I appropriate it.  More eyes are always better.

elishacloud:
> unfortunately std::unordered_set doesn't support indexing. I don't think using std::move is too bad.

@WinterSnowfall, I found a better way to loop through the maps without moving them to a temporary map.

WinterSnowfall:
> @WinterSnowfall, I found a better way to loop through the maps without moving them to a temporary map.

Nicely done. I was actually considering to suggest getting the set keys in a temp vector and iterating over that rather than copying the whole thing, but this is even better.

elishacloud:
@crosire, this is ready to check in.  This fixes a rather prevalent crash that happens on Windows 7.  We are hoping to get this fix into the Silent Hill 2 Enhancements project.

PR #182, opened 2024-10-10
Stub ValidatePixelShader, ValidateVertexShader and DebugSetMute

WinterSnowfall:
There are a few of undocumented functions exposed by d3d8.dll which are (apparently) needed by some games to function properly. Namely:

- `ValidatePixelShader()`, `ValidateVertexShader()` are needed by [Shadow of Memories/Destiny](https://en.wikipedia.org/wiki/Shadow_of_Memories), otherwise the game crashes on startup.
- `DebugSetMute()`, is needed by [C&C: Generals](https://en.wikipedia.org/wiki/Command_%26_Conquer%3A_Generals) (and its addon, Zero Hour) to perform adequently when rendering menus. For some reason the games spam calls to `DebugSetMute()` on the rendering thread (I assume) when menus are on screen (about once every frame) and that causes massive slowdowns if this function isn't exposed.

I've stubbed them for now, and that resolves both issues mentioned above. Potentially, ValidatePixelShader and ValidateVertexShader could be implemented using d3dcompiler (maybe?), but that's a bit beyond what I'm familiar with, so I'll leave it to someone else to pick up the torch if interested.

### DIFF
--- a/res/d3d8.def
+++ b/res/d3d8.def
@@ -1,2 +1,5 @@
 EXPORTS
+	ValidatePixelShader
+	ValidateVertexShader
+	DebugSetMute
 	Direct3DCreate8
--- a/source/d3d8to9.cpp
+++ b/source/d3d8to9.cpp
@@ -15,6 +15,120 @@ PFN_D3DXLoadSurfaceFromSurface D3DXLoadSurfaceFromSurface = nullptr;
 std::ofstream LOG;
 #endif
 
+extern "C" HRESULT WINAPI ValidatePixelShader(const DWORD* pPixelShader, const D3DCAPS8* pCaps, BOOL ReturnErrors, char** pErrorsString)
+{
+#ifndef D3D8TO9NOLOG
+	LOG << "Redirecting '" << "ValidatePixelShader " << "(" << pPixelShader << ", " << pCaps << ", " << ReturnErrors << ", " << pErrorsString << ")' ..." << std::endl;
+#endif
+
+	HRESULT hr = E_FAIL;
+	char* errorMessage = "";
+
+	if (!pPixelShader)
+	{
+		errorMessage = "Invalid code pointer.\n";
+	}
+	else
+	{
+		switch (*pPixelShader)
+		{
+		case D3DPS_VERSION(1, 0):
+		case D3DPS_VERSION(1, 1):
+		case D3DPS_VERSION(1, 2):
+		case D3DPS_VERSION(1, 3):
+		case D3DPS_VERSION(1, 4):
+			if (pCaps && *pPixelShader > pCaps->PixelShaderVersion)
+			{
+				errorMessage = "Shader version not supported by caps.\n";
+				break;
+			}
+			hr = S_OK;
+			break;
+
+		default:
+			errorMessage = "Unsupported shader version.\n";
+		}
+	}
+
+	if (!ReturnErrors)
+	{
+		errorMessage = "";
+	}
+
+	if (pErrorsString)
+	{
+		const size_t size = strlen(errorMessage) + 1;
+
+		*pErrorsString = (char*) HeapAlloc(GetProcessHeap(), 0, size);
+		if (*pErrorsString)
+		{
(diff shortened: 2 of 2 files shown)


elishacloud:
Looks like a partial implementation.  It could have issues.  How about coping the code from [here](https://github.com/elishacloud/dxwrapper/blob/b9c525483c868a685f0bb1364cc6cb7b826e9070/d3d8/d3d8.cpp#L47-L144)?

WinterSnowfall:
> Looks like a partial implementation.

Yes, it's just very basic stubbing which works for the mentioned games, whereas what dxwrapper has is a bit more extensive. Sure, can bring that over, thanks for suggesting it.

elishacloud:
BTW: I copied the code from [Wine](https://gitlab.winehq.org/wine/wine/-/tree/master/dlls/d3d8).

source/d3d8to9.cpp
```diff
@@ -15,6 +15,123 @@ PFN_D3DXLoadSurfaceFromSurface D3DXLoadSurfaceFromSurface = nullptr;
 std::ofstream LOG;
 #endif
 
+extern "C" HRESULT WINAPI ValidatePixelShader(const DWORD* pPixelShader, const D3DCAPS8* pCaps, BOOL ReturnErrors, char** pErrorsString)
+{
+	UNREFERENCED_PARAMETER(pCaps);
+	UNREFERENCED_PARAMETER(ReturnErrors);
```
elishacloud, line 21: I guess you don't need these two lines anymore.

source/d3d8to9.cpp
```diff
@@ -15,6 +15,123 @@ PFN_D3DXLoadSurfaceFromSurface D3DXLoadSurfaceFromSurface = nullptr;
...
+	}
+
+	return hr;
+}
+
+extern "C" HRESULT WINAPI ValidateVertexShader(const DWORD* pVertexShader, const DWORD* pVertexDecl, const D3DCAPS8* pCaps, BOOL ReturnErrors, char** pErrorsString)
+{
+	UNREFERENCED_PARAMETER(pVertexDecl);
+	UNREFERENCED_PARAMETER(pCaps);
+	UNREFERENCED_PARAMETER(ReturnErrors);
```
elishacloud, line 78: I guess you don't need two of these lines anymore.

WinterSnowfall:
Ok, I've carried over the additional validations @elishacloud has graciously linked. They aren't explicitly needed as far as I'm aware (and games are happy if we just return S_OK, really), but I guess the implementation is a bit more complete now.

I would imagine the native/Windows implementation of `ValidateVertexShader()` does more with pVertexDecl in terms of actual validation, but it should suffice for now.

source/d3d8to9.cpp
WinterSnowfall: Yep, caught them after I pushed :+1: .

source/d3d8to9.cpp
WinterSnowfall: All except `pVertexDecl`.

elishacloud:
Looks good!  I have had this in my code for a while and so far I have not found a game that had an issue with them.  So I think we should be good here.

res/d3d8.def
```diff
@@ -1,2 +1,5 @@
 EXPORTS
+	ValidatePixelShader
```
universal963, line 2: I don't know if it is necessary but how about adding ordinal value here?

WinterSnowfall: Ordinals shouldn't be necessary as far as I'm aware (d3d8to9 has worked fine without any so far, and technically Direct3DCreate8 is @ 5 in the native dll, not wherever MSVC has decided to put it until now :), but I can add them if @crosire wants me to.

crosire: IIRC I saw the exports at different ordinals in different Windows versions, indicating that the ordinals are not hardcoded for D3D8 (but may be mixing D3D8 with a different DLL right now), so adding them here would bring no benefit.

PR #183, opened 2024-10-10
Set the correct value of 0.0f for D3DRS_POINTSIZE_MIN

WinterSnowfall:
A minor nitpick, though some games may be affected by it in the end.

The default value of D3DRS_POINTSIZE_MIN in [D3D9 is 1.0f](https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3drenderstatetype#constants), whereas it is 0.0f in D3D8:

image

This needs to be propagated both on device creation and on subsequent device resets.

Besides the documentation hinting at these defaults, I wrote a test to verify them against native d3d8 and d3d9, and they are indeed as stated (thus in need of correction).


crosire:
Could you add a comment with a brief note why the call is made (as described in this PR), so that it's more obvious to future readers of the code?

WinterSnowfall:
Sure, can do.

PR #184, opened 2024-10-16
Add D3DFMT_UNKNOWN checks on object creation

WinterSnowfall:
D3D8 has some specific pointer clearing behavior around D3DFMT_UNKNOWN format validations. These fix several Wine tests, but also [a test I wrote myself](https://github.com/WinterSnowfall/d8vk-tests/commit/059f96ab14854e01c0b2ffa9fb2126cb01e8a91b) that passes on native D3D8.

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -279,6 +279,9 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::CreateTexture(UINT Width, UINT Height
 	if (ppTexture == nullptr)
 		return D3DERR_INVALIDCALL;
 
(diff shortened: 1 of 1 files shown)


ThirteenAG:
May I also ask to fix warnings of this kind while you're at it?

```
d3d8to9.cpp(25,21): error C2440: 'initializing': cannot convert from 'const char [1]' to 'char *' 
```

Project stopped compiling with c++latest after recent merges.

WinterSnowfall:
@ThirteenAG Sorry for the trouble, but it all works fine here in MSVC 2022 (I always compile and check, of course), so not sure what the problem is.

CookiePLMonster:
It's probably a warnings-as-errors thing.

PR #186, opened 2024-10-16
Validate state block calls during state recording

WinterSnowfall:
In D3D8 a call to `BeginStateBlock()` starts state recording and also causes other state block calls (with the exception of `EndStateBlock()`) to fail. D3D9 does not handle this state block call validation behavior automatically, since some calls such as `DeleteStateBlock()` are missing altogether.


elishacloud:
@WinterSnowfall, should you set `IsRecordingState = false;` when `Reset()` is called?

WinterSnowfall:
> @WinterSnowfall, should you set `IsRecordingState = false;` when `Reset()` is called?

D3D8 state blocks are not affected by device resets in any way, so no, I don't think that's warranted.

PR #187, opened 2024-10-16
Fix some char* assignment compiler warnings

WinterSnowfall:
Apparently using non-const char* causes warnings/errors with some C++ compilers, and this *should* be more palatable in general. @ThirteenAG Please let me know if this fixes the problem you were seeing.

### DIFF
--- a/source/d3d8to9.cpp
+++ b/source/d3d8to9.cpp
@@ -22,7 +22,7 @@ extern "C" HRESULT WINAPI ValidatePixelShader(const DWORD* pPixelShader, const D
 #endif
 
 	HRESULT hr = E_FAIL;
-	char* errorMessage = "";
+	const char* errorMessage = "";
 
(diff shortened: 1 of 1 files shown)


CookiePLMonster:
This should just become a `const char*`.

WinterSnowfall:
> This should just become a `const char*`.

We have to modify it based on various validation errors, so that's not really an option.

CookiePLMonster:
> > This should just become a `const char*`.
> 
> We have to modify it based on various validation errors, so that's not really an option.

`const char*` pointer can be reassigned, it just indicates that it points to constant strings.

WinterSnowfall:
> `const char*` pointer can be reassigned, it just indicates that it points to constant strings.

Yeah, you're right of course, I just got a bit confused (my natural state when it comes to C-isms). Should be fine now, thanks for your input.

PR #188, opened 2024-10-16
Setup SWVP toggle for mixed mode devices

WinterSnowfall:
Not entirely sure why this wasn't being done in the first place, as devices created with D3DCREATE_MIXED_VERTEXPROCESSING expect this to work properly and there are some games that require dynamically switching between the two.

In the case of either D3DCREATE_SOFTWARE_VERTEXPROCESSING or D3DCREATE_HARDWARE_VERTEXPROCESSING devices, the call to `SetSoftwareVertexProcessing()` will simply fail, but we can safely ignore that. In fact D3D8 will never return anything other than D3D_OK for any `SetRenderState()` calls.

### DIFF
--- a/source/d3d8to9.hpp
+++ b/source/d3d8to9.hpp
@@ -53,7 +53,7 @@ class Direct3DDevice8 : public IDirect3DDevice8
 	Direct3DDevice8 &operator=(const Direct3DDevice8 &) = delete;
 
 public:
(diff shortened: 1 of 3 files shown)


WinterSnowfall:
I've refactored this to be a bit to be more elegant and not spam D3D9 with calls we know would fail anyway. The end effect is identical.

Issue #194, opened 2025-02-06
SetStreamSource - incompabitility between D3D8 and D3D9


BC46:
The D3D8 documentation mentions the following about `IDirect3DDevice8::SetStreamSource`:
> This method increments the reference count of the vertex buffer being assigned and decrements the reference count of the previously assigned vertex buffer. When the vertex buffer is no longer needed, set it to NULL. If you fail to do this, the vertex buffer is not released, resulting in a memory leak.

This means that `IDirect3DDevice8::SetStreamSource` should be called with `pStreamData = nullptr` to release the vertex buffer. However, if d3d8to9's implementation of this function is called like this, then `D3DERR_INVALIDCALL` is returned.
```cpp
HRESULT STDMETHODCALLTYPE Direct3DDevice8::SetStreamSource(UINT StreamNumber, IDirect3DVertexBuffer8 *pStreamData, UINT Stride)
{
	if (pStreamData == nullptr)
		return D3DERR_INVALIDCALL;
	...
}
```

This is not ideal as for instance in the game Freelancer (2003), every time a D3D8 function returns `D3DERR_INVALIDCALL`, a long string is written to a log file. Since Freelancer calls `IDirect3DDevice8::SetStreamSource` with `pStreamData = nullptr` about every frame, the log file is effectively flooded with this exact string when running the game with d3d8to9. After having the game open for a while, the log file can reach sizes of up to multiple hundreds of MBs. Clearly, this is undesired behavior.

In the D3D9 documentation regarding `IDirect3DDevice9::SetStreamSource` I couldn't find any defined behavior about passing `pStreamData = nullptr` to the function. It seems to not mention anything about a reference count like in the D3D8 documentation either.

WinterSnowfall:
Nice catch. To think I've completed the game once using d3d8to9 and hadn't noticed this...

I agree the call should just be forwarded to d3d9 here, even with a NULL pStreamData. What I've seen that is unique in d3d8 is that on a call with NULL pStreamData the previous/existing Stride is preserved. Not sure if any games actually care about this much, but ideally it'll have to be forwarded with whatever `GetStreamSource()` returns before we pass it along in this case.

BC46:
> Try with this update: [d3d8.zip](https://github.com/user-attachments/files/18697748/d3d8.zip)

[...]

PR #195, opened 2025-02-06
Fix SetStreamSource with nullptr

elishacloud:
See comments from #194

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -1513,7 +1513,8 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::SetVertexShader(DWORD Handle)
 		hr = ProxyInterface->SetVertexShader(ShaderInfo->Shader);
 		ProxyInterface->SetVertexDeclaration(ShaderInfo->Declaration);
 
-		CurrentVertexShaderHandle = Handle;
+		if (SUCCEEDED(hr))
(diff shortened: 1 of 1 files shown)


crosire:
This probably should apply to `IDirect3DDevice8::SetIndices` too, it doesn't fail when passing nullptr in D3D9 at least. Looks like in both cases D3D9 doesn't actually unbind the current buffers when passed `nullptr` ... and I suspect it behaves the same in D3D8 and the documentation just isn't very helpful here.

WinterSnowfall:
> Looks like in both cases D3D9 doesn't actually unbind the current buffers when passed `nullptr`

There is a d3d9 game that depends on this particularity actually (buffers remaining unbound), namely Operation Flashpoint: Red River. I don't know any d3d8 games that are affected, but I also think it's safe to assume the behavior is similar here.

elishacloud:
> This probably should apply to `IDirect3DDevice8::SetIndices` too, it doesn't fail when passing nullptr in D3D9 at least.

Good idea.  I put an update in for SetIndices.  I also changed the formatting to try and better match the rest of the code.

elishacloud:
I put one more change in to check the return value before setting the current handle.  Also, I fixed some additional formatting changes to mirror the rest of the code.

@crosire, I'm done making changes.  If you are ok with the changes you can merge them.

PR #68, opened 2018-03-17
Fix Multisample support

elishacloud:
This fixes #66 and #67.

**Overview:**

There are three updates in this fix:

1. Updated CreateDevice, CreateAdditionalSwapChain, CreateRenderTarget, CreateDepthStencilSurface and Reset functions to support MultiSampleQuality by querying CheckDeviceMultiSampleType.  If MultiSampleType == D3DMULTISAMPLE_NONE then set MultiSampleQuality to 0.  Note: this is all the functions that use MultiSampling as seen [here](https://msdn.microsoft.com/en-us/library/windows/desktop/bb172574(v=vs.85).aspx).
2. Remove D3DPRESENTFLAG_LOCKABLE_BACKBUFFER flag when MultiSampleType != D3DMULTISAMPLE_NONE is detected.
3. Update unsupported D3DPRESENT_RATE_UNLIMITED flag to use D3DPRESENT_INTERVAL_IMMEDIATE rather than D3DPRESENT_INTERVAL_ONE to replicate more closely the d3d8 functionality.

**Testing:**

Tested the following games:

- Codemasters Race Driver 2
- Grand Theft Auto 3
- Grand Theft Auto Vice City
- Haegemonia Legions of Iron
- Hitman 2 Silent Assassin
- Indiana Jones and the Emperor's Tomb
- Legacy of Kain Blood Omen 2
- Max Payne
- Max Payne 2
- Moto Racer 3
- Need for Speed III
- Need For Speed Hot Pursuit 2
- Raymond 3
- Serious Sam The First Encounter
- Serious Sam The Second Encounter
- Silent Hill 2
- Star Wars Republic Commando
- Star Wars Starfighter
- True Crime New York City

PR #196, opened 2025-03-21
Default to using 0 for MultiSampleQuality

WinterSnowfall:
I'm not entirely sure why the existing logic was set up in the first place.

The MultiSampleQuality parameter in D3D9 is only intended for use with `D3DMULTISAMPLE_NONMASKABLE`, which is not available in D3D8.

This is implied by the official documentation as well:

> D3DMULTISAMPLE_NONMASKABLE
>     Enables the multisample quality value. See Remarks.

> Whether the display device supports maskable multisampling (more than one sample for a multiple-sample render-target format plus antialias support) or just non-maskable multisampling (only antialias support), the driver for the device provides the number of quality levels for the D3DMULTISAMPLE_NONMASKABLE multiple-sample type.

Therefore, while the current code doesn't do any harm, it is rather pointlessly issuing redundant queries to D3D9, as with anything other than `D3DMULTISAMPLE_NONMASKABLE`, `CheckDeviceMultiSampleType` will return `1` for the number of quality levels.

I'm linking the WineD3D implementation of `CheckDeviceMultiSampleType` below, but I have no reason to believe native behaves any differently in modern times, nor that it did back in the day:

https://github.com/wine-mirror/wine/blob/3323a7e4474afc3509ba56eab51b76e7887f6450/dlls/wined3d/directx.c#L1925-L1935

On a side-note, `D3DMULTISAMPLE_NONMASKABLE` is not frequently used on D3D9 side either, but the games that do employ it do indeed set a non-zero MultiSampleQuality value, while all the other games that resort to maskable multisamples use 0.

In case anyone wants a reference for `D3DMULTISAMPLE_NONMASKABLE` use in the future, the games I am aware of (9 out of 125 D3D9 titles I've looked into) are:
- Star Wars: Battlefront
- AquaNox 2: Revelation
- Command & Conquer 3: Tiberium Wars
- Dreamfall: The Longest Journey
- El Matador
- Heroes of Might and Magic V
- Machinarium
- Need for Speed: Underground 2
- The Void

### DIFF
--- a/source/d3d8to9_base.cpp
+++ b/source/d3d8to9_base.cpp
@@ -180,21 +180,6 @@ HRESULT STDMETHODCALLTYPE Direct3D8::CreateDevice(UINT Adapter, D3DDEVTYPE Devic
 	D3DPRESENT_PARAMETERS PresentParams;
 	ConvertPresentParameters(*pPresentationParameters, PresentParams);
 
-	// Get multisample quality level
-	if (PresentParams.MultiSampleType != D3DMULTISAMPLE_NONE)
-	{
-		DWORD QualityLevels = 0;
-		if (ProxyInterface->CheckDeviceMultiSampleType(Adapter,
-			DeviceType, PresentParams.BackBufferFormat, PresentParams.Windowed,
-			PresentParams.MultiSampleType, &QualityLevels) == S_OK &&
-			ProxyInterface->CheckDeviceMultiSampleType(Adapter,
-				DeviceType, PresentParams.AutoDepthStencilFormat, PresentParams.Windowed,
-				PresentParams.MultiSampleType, &QualityLevels) == S_OK)
-		{
-			PresentParams.MultiSampleQuality = (QualityLevels != 0) ? QualityLevels - 1 : 0;
-		}
-	}
-
 	IDirect3DDevice9 *DeviceInterface = nullptr;
 
 	const HRESULT hr = ProxyInterface->CreateDevice(Adapter, DeviceType, hFocusWindow, BehaviorFlags, &PresentParams, &DeviceInterface);
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -169,24 +169,6 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::CreateAdditionalSwapChain(D3DPRESENT_
 	D3DPRESENT_PARAMETERS PresentParams;
 	ConvertPresentParameters(*pPresentationParameters, PresentParams);
 
-	// Get multisample quality level
-	if (PresentParams.MultiSampleType != D3DMULTISAMPLE_NONE)
-	{
-		DWORD QualityLevels = 0;
-		D3DDEVICE_CREATION_PARAMETERS CreationParams;
-		ProxyInterface->GetCreationParameters(&CreationParams);
-
-		if (D3D->GetProxyInterface()->CheckDeviceMultiSampleType(CreationParams.AdapterOrdinal,
-			CreationParams.DeviceType, PresentParams.BackBufferFormat, PresentParams.Windowed,
-			PresentParams.MultiSampleType, &QualityLevels) == S_OK &&
-			D3D->GetProxyInterface()->CheckDeviceMultiSampleType(CreationParams.AdapterOrdinal,
-				CreationParams.DeviceType, PresentParams.AutoDepthStencilFormat, PresentParams.Windowed,
-				PresentParams.MultiSampleType, &QualityLevels) == S_OK)
-		{
-			PresentParams.MultiSampleQuality = (QualityLevels != 0) ? QualityLevels - 1 : 0;
-		}
-	}
-
 	IDirect3DSwapChain9 *SwapChainInterface = nullptr;
 
 	const HRESULT hr = ProxyInterface->CreateAdditionalSwapChain(&PresentParams, &SwapChainInterface);
@@ -210,24 +192,6 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::Reset(D3DPRESENT_PARAMETERS8 *pPresen
 
 	D3DPRESENT_PARAMETERS PresentParams;
 	ConvertPresentParameters(*pPresentationParameters, PresentParams);
-
-	// Get multisample quality level
-	if (PresentParams.MultiSampleType != D3DMULTISAMPLE_NONE)
-	{
-		DWORD QualityLevels = 0;
-		D3DDEVICE_CREATION_PARAMETERS CreationParams;
-		ProxyInterface->GetCreationParameters(&CreationParams);
-
-		if (D3D->GetProxyInterface()->CheckDeviceMultiSampleType(CreationParams.AdapterOrdinal,
-			CreationParams.DeviceType, PresentParams.BackBufferFormat, PresentParams.Windowed,
-			PresentParams.MultiSampleType, &QualityLevels) == S_OK &&
-			D3D->GetProxyInterface()->CheckDeviceMultiSampleType(CreationParams.AdapterOrdinal,
-				CreationParams.DeviceType, PresentParams.AutoDepthStencilFormat, PresentParams.Windowed,
-				PresentParams.MultiSampleType, &QualityLevels) == S_OK)
-		{
-			PresentParams.MultiSampleQuality = (QualityLevels != 0) ? QualityLevels - 1 : 0;
-		}
-	}
 
 	HRESULT hr = ProxyInterface->Reset(&PresentParams);
 
@@ -395,21 +359,9 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::CreateRenderTarget(UINT Width, UINT H
 
 	*ppSurface = nullptr;
 
-	DWORD QualityLevels = 0;
-
-	// Get multisample quality level
-	if (MultiSample != D3DMULTISAMPLE_NONE)
-	{
-		D3DDEVICE_CREATION_PARAMETERS CreationParams;
-		ProxyInterface->GetCreationParameters(&CreationParams);
-
-		D3D->GetProxyInterface()->CheckDeviceMultiSampleType(CreationParams.AdapterOrdinal, CreationParams.DeviceType, Format, FALSE, MultiSample, &QualityLevels);
-		QualityLevels = (QualityLevels != 0) ? QualityLevels - 1 : 0;
-	}
-
 	IDirect3DSurface9 *SurfaceInterface = nullptr;
 
-	const HRESULT hr = ProxyInterface->CreateRenderTarget(Width, Height, Format, MultiSample, QualityLevels, Lockable, &SurfaceInterface, nullptr);
+	const HRESULT hr = ProxyInterface->CreateRenderTarget(Width, Height, Format, MultiSample, 0, Lockable, &SurfaceInterface, nullptr);
 	if (FAILED(hr))
 		return hr;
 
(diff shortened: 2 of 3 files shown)


WinterSnowfall:
Oh, I see. If that is indeed the intention here, I'm fine to close the PR. But from a "barebones" D3D8 perspective, none of this makes any sense.

HerMajestyDrMona:
You're probably right about this change to reproduce the original behavior. I just wanted to bring up this possibility, as I myself use something like this:

Screenshot 2025-03-21 173339

And I am not sure if `MultiSampleQuality` has any effect or not? It should have if it's the graphics driver deciding about it.

WinterSnowfall:
> And I am not sure if `MultiSampleQuality` has any effect or not? It should have if it's the graphics driver deciding about it.

It would be the driver indeed. The D3D9 documentation is somewhat esoteric about what these do:

> The meanings of these quality levels are defined by the device manufacturer and cannot be queried through D3D. For example, for a particular device different quality levels at a fixed sample count might refer to different spatial layouts of the sample locations or different methods of resolving.

crosire:
This was added in #68. The wording in the documentation is a bit difficult to decypher the meaning from, but it seems reasonable to remove. The crashes fixed by #68 were related to the `D3DPRESENTFLAG_LOCKABLE_BACKBUFFER` flag, so shouldn't be affected.

WinterSnowfall:
> When I put that code in I was trying to get the best quality I could, not realizing that it only affected `D3DMULTISAMPLE_NONMASKABLE`.

You might have also run into this: https://learn.microsoft.com/en-us/windows/win32/direct3d9/converting-to-directx-9#multisampling-quality-changes , which sort of hints at the idea one should query for the number of quality levels anyway. I find that section even more theoretical (and confusing) than the official d3d9 documentation because in practice I have not seen any quality level other than 0 on maskable multisample types. Such fine grained quality adjustments might simply not have been done by drivers or it might be a case of conflicting docs, which d3d is not a stranger to.

WinterSnowfall:
> If I understand the discussion correctly, in short, for regular full-scene antialiasing the type would be e.g. `D3DMULTISAMPLE_8_SAMPLES` for MSAA 8x, and the qualitylevel field does nothing as it's only meant for `D3DMULTISAMPLE_NONMASKABLE`?

In theory, the discussion is more nuanced, as I've remarked above:

> The meanings of these quality levels are defined by the device manufacturer and cannot be queried through D3D. For example, for a particular device different quality levels at a fixed sample count might refer to different spatial layouts of the sample locations or different methods of resolving.

So `D3DMULTISAMPLE_8_SAMPLES` *could* also have multiple quality levels depending on the vendor implementation according to the spec, but in practice I don't think this was ever used.

So essentially, yes, that is the observed behavior for D3D9. D3D8, however, had no concept of these fine grained quality levels to begin with, so defaulting to 0 certainly makes sense.

anzz1:
Now I am confused, wasn't the premise of the PR that the MultiSampleQuality is ever applicable to D3DMULTISAMPLE_NONMASKABLE, as even stated in the commit?
`// MultiSampleQuality is only used in conjunction with D3DMULTISAMPLE_NONMASKABLE, which is not available in D3D8`
https://github.com/crosire/d3d8to9/pull/196/files#diff-94bd12234525af0e274efc25a73accf3c9c690e32431026851596df10e939bbfR135

If there indeed is a chance that the other modes could also have quality levels, wasn't the implementation *before* already correct? That is, default to the highest quality instead of the lowest?
Those queries was already only used in the case multisampling was used.

Quick google search shows that D3D9 implementations look often something like this:
```c
DWORD MSQuality = 0;
D3DMULTISAMPLE_TYPE MSType = D3DMULTISAMPLE_NONE;
if( SUCCEEDED( m_pD3D->CheckDeviceMultiSampleType(D3DADAPTER_DEFAULT, D3DDEVTYPE_HAL, D3DFMT_A8R8G8B8, TRUE, D3DMULTISAMPLE_8_SAMPLES, &MSQuality) ))
	MSType = D3DMULTISAMPLE_8_SAMPLES;
else if( SUCCEEDED( m_pD3D->CheckDeviceMultiSampleType(D3DADAPTER_DEFAULT, D3DDEVTYPE_HAL, D3DFMT_A8R8G8B8, TRUE, D3DMULTISAMPLE_4_SAMPLES, &MSQuality) ))
	MSType = D3DMULTISAMPLE_4_SAMPLES;
else if( SUCCEEDED( m_pD3D->CheckDeviceMultiSampleType(D3DADAPTER_DEFAULT, D3DDEVTYPE_HAL, D3DFMT_A8R8G8B8, TRUE, D3DMULTISAMPLE_2_SAMPLES, &MSQuality) ))
	MSType = D3DMULTISAMPLE_2_SAMPLES;
int MSQ = MSQuality - 1;
...
```
Which would be setting the MultiSampleQuality field to highest of whatever the driver reports.

To me it seems that the implementation before this commit was the correct way to do it. And I understand the point, I am too a fan of making everything as tight as possible, but it seems that in this case having the calls there was the proper way of implementation, and the performance impact of them is near-nil.

WinterSnowfall:
> Now I am confused, wasn't the premise of the PR that the MultiSampleQuality is ever applicable to D3DMULTISAMPLE_NONMASKABLE, as even stated in the commit?

In practice I have not found any prof of this being false, I was only stating that the docs state support for such mythical beasts exist in theory.

> If there indeed is a chance that the other modes could also have quality levels, wasn't the implementation _before_ already correct?

I have looked at quite a few native D3D9 traces and found no use of non-0 quality levels outside of `D3DMULTISAMPLE_NONMASKABLE`. There is a chance it *might* have been different in ancient drivers (a known unknown), but I've found no evidence of modern drivers exposing multiple maskable quality levels.

And again, D3D8 has no concept of quality levels, so if you want an accurate reproduction of how D3D8 looked, 0 is the way to go, as the "default" quality was the only option implicitly available. Reverse engineering D3D9 type queries to pick anything else would qualify as an "enhancement" even if multiple quality levels were ever exposed, ad absurdum.

anzz1:
> Run that code against modern drivers and you'll find MSQ is always 0.

Right, this was my point if this is indeed true for drivers made during the D3D9 lifespan of 23 years as of writing then it indeed does not matter.  But is it really?

> I have looked at quite a few native D3D9 traces and found no use of non-0 quality levels outside of D3DMULTISAMPLE_NONMASKABLE.

This was where I was afraid of you having an error in your methodology, if you check only the apitrace of different *applications* instead of different *drivers*, that wouldn't yield in result of any probative value as you can see any code that uses the return value from the `CheckMultiSampleType` for the `MultiSampleQuality`, then that value on apitrace would be dependent on the *driver* used, not the *application*.

>so if you want an accurate reproduction of how D3D8 looked, 0 is the way to go, as the "default" quality was the only option implicitly available.

I couldn't find anything backing the claim of 0 being the comparable "default" value, if anything one would presume it'd be a middle "medium quality" value if available. And in real-world use cases, which often are like @HerMajestyDrMona already pointed out adding additional quality improvements (why else even bother upgrading from 8 -> 9), I don't see many scenarios where the user would want their mipmapping quality lowered.

> Run that code against modern drivers and you'll find MSQ is always 0.

D3D8 and D3D9 are both ancient technologies (8 released in 2000, and 9 released in 2002) and they're no doubt already heavily "shimmed" in modern Windows and modern drivers. If that was everything you've tested against, that's definitely wrong. Testing against an array of period OS and period drivers would be warranted before making any sort of call about what the drivers return in various scenarios and what not.

Ultimately the problem I'm seeing here is *why*, if a thorough investigation that this isn't actually used anywhere isn't done, and change it from the way D3D9 is usually implemented querying the mipmap quality levels from the driver, there should be at least some reason to make it, something positive it brings, but I am only seeing a potential negative without any positive.

WinterSnowfall:
> But is it really?

Feel free to test as extensively as you wish and should you find anything different in the wild, then PR something to revert the change. My intention was to get rid of superfluous code based on observed behavior of modern drivers, which is what the large majority of people will be using these days.

> This was where I was afraid of you having an error in your methodology.

I have looked at traces collected on a variety of hardware: Intel, AMD and Nvidia.

> I couldn't find anything backing the claim of 0 being the comparable "default" value, if anything one would presume it'd be a middle "medium quality" value if available.

I try to presume as little as I can. 0 in this case is not only a default but the only quality level drivers expose in practice.

> Ultimately the problem I'm seeing here is _why_, if a thorough investigation that this isn't actually used anywhere isn't done, and change it from the way D3D9 is usually implemented querying the mipmap quality levels from the driver, there should be at least some reason to make it, something positive it brings, but I am only seeing a potential negative without any positive.

I think I've explained the reasons for this change/simplification as best I could, but if you find my explanation unsatisfactory, as I mentioned above, feel free to PR your own changes based on your reasoning. I do not own this repository, I simply contribute to it. This isn't in any way related to mipmap quality by the way, not sure why you're bringing that up.

anzz1:
Alright, it seems you have considered this through. It seems that there is no mistakes here, and you understood that the apitrace in this case relates to the driver, not the application (for applications using reference implementation of using the return value as quality). While I might not agree with the conclusion made, it was not made in error and due diligence was done, which is the important part.

WinterSnowfall:
> Right, as long as it includes older OS (NT kernel 5.1/5.2/6.0/6.1)/drivers/graphic cards and not just the newest ones on newest OS, where old Direct3D versions are heavily shimmed as mentioned before, that would be proper diligence. Retro gaming after all is while not the only application for this, I'd hazard a guess that it's by far the most used.

I'm not sure I get where you're coming from. There's little to no reason to consider the behavior of anything but modern drivers.

I have a Windows XP based retro PC and a variety of AGP cards from the mid 2000s, but again, these tests will be done out of curiosity more than anything else. You gain absolutely nothing by using d3d8to9 on these types of setups, because D3D8 support is stellar to begin with. Its main utility these days is on modern (aka still in support) Windows OSes where D3D8 is indeed, as you say, "shimmed" on top of d3d9, which sometimes breaks compatibility with older games.

Actually any game using d3d earlier than 9 is in the same boat really, which is why you see various implementations these days for ddraw/D3D<8 on top of D3D9 shipped with commercial games (I am thinking mostly of what GOG is doing, but it's not a singular example).

WinterSnowfall:
While looking into other things, I stumbled into an interesting piece of history: looks like on native drivers from the 2010+ years, both nonmaskable and fixed sample quality levels were used to expose support for CSAA in the case of Nvidia and EQAA in the case of AMD. Clearly, both of these by far postdate the d3d8 era, so would not have been available at the time.

Quality level 0 always gets you the corresponding good ol' classic MSAA level, which is what d3d8 would have used anyway.

https://www.geeks3d.com/20110915/amd-eqaa-modes-enhanced-quality-anti-aliasing-for-radeon-hd-6900-series/

That being said, I still stand by what this PR has done and consider it the correct approach. Should you want to revert the change to potentially make use of CSAA/EQAA, at least on certain drivers/cards, I wouldn't mind it terribly either, but it's just not age accurate, and I'm not entirely sure it's worth it either. CSAA/EQAA end up in the land of diminishing returns IMHO at 8x, which is what most people will use anyway in this day and age.

WinterSnowfall:
Alright, finally wrote a test for CheckDeviceMultiSampleType and ran it on d3d8-era hardware / Windows XP... and the results are somewhat unexpected.

On a GeForce GeForce4 Ti 4800 SE, D3D8 reports support for the following types of multisample:

  - D3DMULTISAMPLE_NONE
  - D3DMULTISAMPLE_2_SAMPLES
  - D3DMULTISAMPLE_3_SAMPLES
  - D3DMULTISAMPLE_4_SAMPLES

D3D9, however, on the same hardware/driver combination reports support for:

  - D3DMULTISAMPLE_NONE (Quality levels: 1)
  - D3DMULTISAMPLE_NONMASKABLE (Quality levels: 4)
 
I guess maskable sample support came in a bit later, and all samples in D3D8 were actually non-maskable anyway? I don't have any "newer" retro GPUs that might include hardware support for CSAA/EQAA, sadly, would have been curious to see the situation there.

PR #200, opened 2025-07-06
Additional checks for StateBlock set

elishacloud:
This pull request does two things:

1. It clears the StateBlock cache on Reset() because the interfaces are no longer available after a Reset.
2. It checks the StateBlock cache to validate the handle before accessing, to prevent a crash in case a game sends in an invalid or released handle.

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -195,6 +195,9 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::Reset(D3DPRESENT_PARAMETERS8 *pPresen
 
 	HRESULT hr = ProxyInterface->Reset(&PresentParams);
 
+	// StateBlocks are destroyed during a Reset in D3D9 so need to clear the cache to handle this properly
+	StateBlockTokens.clear();
+
 	if (SUCCEEDED(hr))
 	{
 		// The default value of D3DRS_POINTSIZE_MIN is 0.0f in D3D8,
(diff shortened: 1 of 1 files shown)


source/d3d8to9_device.cpp
```diff
@@ -195,6 +195,9 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::Reset(D3DPRESENT_PARAMETERS8 *pPresen
 
 	HRESULT hr = ProxyInterface->Reset(&PresentParams);
 
+	// StateBlocks are destroyed during a Reset in D3D9 so need to clear the cache to handle this properly
+	StateBlockTokens.clear();
```
WinterSnowfall, line 199: As mentioned above, this should not be done.

WinterSnowfall (changes requested):
As mentioned before

elishacloud:
> Feel free to merge it, it may possibly be fine, however I don't think it will pan out better than what is currently in place.

You're right.  I missed the fact that the interface is still available after the Reset.  I will close this PR,  Thanks!

WinterSnowfall:
> I thought you said that "using an invalid token will crash the d3d8 native runtime".

Indeed it does crash the native runtime whenever that happens. Not all D3D8 era games were that stable to begin with :).

elishacloud:
Thanks for checking this, @WinterSnowfall!

WinterSnowfall:
No worries. I (unfortunately) happen to be painfully familiar with how state blocks work in D3D8 vs D3D9, because I had the honor of checking all the WineD3D state block tests against d8vk, and then scratching my head on how to go about fixing them. Thankfully that's now done and dusted.

PR #201, opened 2025-07-06
Check interface address cache on creation

elishacloud:
This pull request replaces the "new" call to create an interface class with the `FindAddress` call.  This way it can use an existing interface that has the same ProxyInterface address (if it exists) instead of creating a new one.

Currently, d3d8to9 will only delete interface classes when the root device is released.  If a game creates new interfaces all the time (like StateBlocks or Textures) it could use a lot of memory before the device is released.  This update might help in some cases where the device creates interfaces with the same address as a previous interface (of the same type) to attempt to reduce the memory usage.

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -175,7 +175,7 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::CreateAdditionalSwapChain(D3DPRESENT_
 	if (FAILED(hr))
 		return hr;
 
-	*ppSwapChain = new Direct3DSwapChain8(this, SwapChainInterface);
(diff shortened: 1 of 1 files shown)


WinterSnowfall:
In case you were worried about this in particular, these changes should be fine in terms of state block handling. Even if certain resources captured in state blocks are released on Reset, that would/can happen in the native runtime as well, so "corrupted" state blocks are a possibility and it's expected applications won't manage them improperly.

Truly persistent state blocks in D3D8 (those that applications are expected to use even after device resets) will usually only record changes to render states and the like. Yes, it's a bit of a mess, which is why it was changed/addressed in D3D9.

PR #203, opened 2025-07-06
Validate StateBlock Tokens

elishacloud:
This pull request checks the StateBlock cache to validate the handle before accessing, to prevent a crash in case a game sends in an invalid or released handle.


elishacloud:
Change this to return D3D_OK in this case.  See: https://github.com/doitsujin/dxvk/blob/daed0c1ce8d39e6dcc1580b753554deb7fcbd2ae/src/d3d8/d3d8_device.cpp#L1311-L1314

WinterSnowfall (approved):

Issue #189, opened 2024-11-03
 A.I.M., Механоиды main menu not drawn


lukasva77:
I have discovered your wrapper due to gta vice city makes everything better more fps and stutter is gone but invisible bink video but it seems to be a known issue

so I've gone looking in my library for more dx8 games

Advanced Battlegrounds: The Future of Combat or Chrome: Gold Edition also increases fps and works fine I checked just the opening of the first level but whatever

but the game in the title fails in the main menu

just reporting if I find more games I will open more issues

WinterSnowfall:
> but the game in the title fails in the main menu

I've checked the game and it seems to work fine, as long as you don't open the Options pop-up from the main menu. Unfortunately, that relies on the device being lost during window focus loss to work properly which may not behave properly on modern Windows, but as long as you set the options beforehand (using `SetOptions.exe`) and don't open the Options from the main menu, you shouldn't run into any issues.

WinterSnowfall:
> Tested this a bit so the problem occurs from changing the graphics setting to anisotropic filtering default is trilinear
> 
> I didn't think that graphic options could do this but WinterSnowfall saying it works fine I had to check

Strange. I have it set to AF x16 and it's fine.

WinterSnowfall:
> also a really useless menu item cause can just run SetOptions.exe but whatever

Agreed it's not a great design. But I think I know what might be causing the crash - at least I have some ideas to try.

As to the anisotropy setting, not sure what's going on there, as I can't reproduce the problem.

PR #204, opened 2025-07-09
Release state blocks on Reset if the device has been lost

WinterSnowfall:
@elishacloud Looks like you were onto something with #200, but I suggest we limit its scope a bit for the reasons I've explained in that PR.

Regarding #189, the problem with the game crashing after a return from the "Options" window pop-up is because the game relies on device lost behavior, as I suspected. However, that is actually signaled properly by d3d9, but the reset fails because the game doesn't delete its previously created state blocks before that happens.

The solution below isn't entirely ideal either, however in such cases it makes sense because the application has no alternative but to call Reset (most common) or release and recreate another device once its current device becomes lost, and if it can't do that, well... it will crash or hang, as observed with A.I.M.

As to the handling I've done below, excerpt from the docs: 

> If an application detects a lost device, it should pause and periodically call IDirect3DDevice9::TestCooperativeLevel until it receives a return value of D3DERR_DEVICENOTRESET. The application may then attempt to reset the device by calling [IDirect3DDevice9::Reset](https://learn.microsoft.com/en-us/windows/desktop/api/d3d9/nf-d3d9-idirect3ddevice9-reset) and, if this succeeds, restore the necessary resources and resume normal operation.

This change is enough to get things working properly in A.I.M. at least, with no ill side effects. I suspect it might also help in other situations when games crash or get stuck while alt+tab-ing or have a regular flow in which the main game window loses focus. To be a bit more robust in handling these cases, I suggest #203 gets merged before this PR, because then at least we would not crash if a game attempts to reuse (post-reset) a token for a state block we were forced to release.

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -190,10 +190,20 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::Reset(D3DPRESENT_PARAMETERS8 *pPresen
 
 	pCurrentRenderTarget = nullptr;
 
+	const HRESULT deviceState = ProxyInterface->TestCooperativeLevel();
+
+	if (deviceState == D3DERR_DEVICENOTRESET) {
+		while (!StateBlockTokens.empty())
+		{
+			DWORD Token = *StateBlockTokens.begin();
+			DeleteStateBlock(Token);
+		}
+	}
+
 	D3DPRESENT_PARAMETERS PresentParams;
 	ConvertPresentParameters(*pPresentationParameters, PresentParams);
 
-	HRESULT hr = ProxyInterface->Reset(&PresentParams);
+	const HRESULT hr = ProxyInterface->Reset(&PresentParams);
 
 	if (SUCCEEDED(hr))
 	{


source/d3d8to9_device.cpp
```diff
@@ -190,10 +190,20 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::Reset(D3DPRESENT_PARAMETERS8 *pPresen
 
 	pCurrentRenderTarget = nullptr;
 
+	const HRESULT deviceState = ProxyInterface->TestCooperativeLevel();
+
+	if (deviceState == D3DERR_DEVICENOTRESET) {
```
elishacloud, line 195: @WinterSnowfall  I was thinking about making this exact change.  Do you think it makes sense to change this to:

```c++
if (deviceState != D3D_OK) {
```

Is there any other state besides OK that we would want to keep the state blocks?

WinterSnowfall: I've kept it only on `D3DERR_DEVICENOTRESET` because that's the only state that a lost device will signal before a (potentially successful) reset call is issued. If any other error code is returned, there's a high probability the Reset call will fail anyway, and some games may actually rely on that kind of behavior.

So, in summary, we should only care about the optimistic scenario here, which is solely signaled by `D3DERR_DEVICENOTRESET`.

PR #206, opened 2025-08-01
Rayman 3: Protect against Use-After-Release in DeleteVertexShader() and DeletePixelShader()

ProgrammierPatrick:
This PR resolves a crash in Rayman 3 that occurs sometimes during loading screens. The issue was traced to the `IDirect3DDevice::DeleteVertexShader(0x17369F8C)` function, specifically within the following code:

```cpp
const DWORD HandleMagic = Handle << 1;
VertexShaderInfo *const ShaderInfo = reinterpret_cast<VertexShaderInfo *>(HandleMagic);

if (ShaderInfo->Shader != nullptr) 
{
    ShaderInfo->Shader->Release(); // crashed here
    VertexShaderAndDeclarationCount--;
}
```

The crash happened because `ShaderInfo->Shader` was not `nullptr`, but its vtable pointer had already been nulled out. This probably means that the game called `DeleteVertexShader()` twice on the shader.

I managed to solve this issue by checking if the handle is still part of `VertexShaderHandles`. If not,  <del>`D3D_ERRINVALIDCALL `</del> `D3D_OK` is returned, and no additional `Release()`-calls are executed. This change was applied to both `DeleteVertexShader()` and `DeleteVertexShader()`.

This issue is similar to the one addressed in PR #203. Like described in https://github.com/crosire/d3d8to9/pull/200#issuecomment-3042800958, I changed the returned Value to `D3D_OK`.

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -1639,11 +1639,8 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::DeleteVertexShader(DWORD Handle)
 	if ((Handle & 0x80000000) == 0)
 		return D3DERR_INVALIDCALL;
 
-	auto it = VertexShaderHandles.find(Handle);
-	if (it == VertexShaderHandles.end())
-		return D3D_OK;
-
-	VertexShaderHandles.erase(it);
+	if (VertexShaderHandles.erase(Handle) == 0)
+		return D3DERR_INVALIDCALL;
 
 	if (CurrentVertexShaderHandle == Handle)
 	{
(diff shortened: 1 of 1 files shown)


ProgrammierPatrick:
Rayman 3 is Great 😄 The game does not care for the returned value. This is the function within Rayman3.exe where the crash happened:
grafik

Initially I tested my code with `D3DERR_INVALIDCALL`, which also worked fine.

Wine apparently always returnes `D3DERR_INVALIDCALL`, even in DeleteStateBlock()
- https://gitlab.winehq.org/wine/wine/-/blob/master/dlls/d3d8/device.c#L3012
- https://gitlab.winehq.org/wine/wine/-/blob/master/dlls/d3d8/device.c#L2182

DXVK also agrees to return `D3DERR_INVALIDCALL` in DeleteVertexShader:
- https://github.com/doitsujin/dxvk/blob/daed0c1ce8d39e6dcc1580b753554deb7fcbd2ae/src/d3d8/d3d8_device.cpp#L2004

I went ahead and changed this PR to return `D3DERR_INVALIDCALL` again.

PR #209, opened 2025-10-19
Better handling of ZBias to DepthBias computing

elishacloud:
This pull request updates the conversion from ZBias to DepthBias, taking into account the actual depth buffer bit count in use.

This ensures a more accurate and consistent translation between the two bias systems.

Additionally, it guarantees that `GetRenderState(D3DRS_ZBIAS)` now always returns the correct corresponding value.

### DIFF
--- a/source/d3d8to9.hpp
+++ b/source/d3d8to9.hpp
@@ -53,7 +53,7 @@ class Direct3DDevice8 : public IDirect3DDevice8
 	Direct3DDevice8 &operator=(const Direct3DDevice8 &) = delete;
 
 public:
-	Direct3DDevice8(Direct3D8 *d3d, IDirect3DDevice9 *ProxyInterface, DWORD BehaviorFlags, BOOL EnableZBufferDiscarding = FALSE);
+	Direct3DDevice8(Direct3D8 *d3d, IDirect3DDevice9 *ProxyInterface, DWORD BehaviorFlags, D3DFORMAT ZBufferFormat, BOOL EnableZBufferDiscarding);
 	~Direct3DDevice8();
 
 	IDirect3DDevice9 *GetProxyInterface() const { return ProxyInterface; }
@@ -165,8 +165,10 @@ private:
 
 	Direct3D8 *const D3D;
 	IDirect3DDevice9 *const ProxyInterface;
-	INT CurrentBaseVertexIndex = 0;
 	const BOOL ZBufferDiscarding = FALSE;
+	DWORD ZBufferBitCount = 0;
+	DWORD ZBiasRenderState = 0;
+	INT CurrentBaseVertexIndex = 0;
 	DWORD CurrentVertexShaderHandle = 0, CurrentPixelShaderHandle = 0;
 	IDirect3DSurface9 *pCurrentRenderTarget = nullptr;
 	bool PaletteFlag = false;
--- a/source/d3d8to9_base.cpp
+++ b/source/d3d8to9_base.cpp
@@ -186,7 +186,7 @@ HRESULT STDMETHODCALLTYPE Direct3D8::CreateDevice(UINT Adapter, D3DDEVTYPE Devic
 	if (FAILED(hr))
 		return hr;
 
-	*ppReturnedDeviceInterface = new Direct3DDevice8(this, DeviceInterface, BehaviorFlags, (PresentParams.Flags & D3DPRESENTFLAG_DISCARD_DEPTHSTENCIL) != 0);
+	*ppReturnedDeviceInterface = new Direct3DDevice8(this, DeviceInterface, BehaviorFlags, PresentParams.AutoDepthStencilFormat ? PresentParams.AutoDepthStencilFormat : D3DFMT_UNKNOWN, (PresentParams.Flags & D3DPRESENTFLAG_DISCARD_DEPTHSTENCIL) != 0);
 
 	// Set default vertex declaration
 	DeviceInterface->SetFVF(D3DFVF_XYZ);
--- a/source/d3d8to9_base.cpp
+++ b/source/d3d8to9_base.cpp
@@ -186,7 +186,7 @@ HRESULT STDMETHODCALLTYPE Direct3D8::CreateDevice(UINT Adapter, D3DDEVTYPE Devic
 	if (FAILED(hr))
 		return hr;
 
-	*ppReturnedDeviceInterface = new Direct3DDevice8(this, DeviceInterface, BehaviorFlags, PresentParams.AutoDepthStencilFormat ? PresentParams.AutoDepthStencilFormat : D3DFMT_UNKNOWN, (PresentParams.Flags & D3DPRESENTFLAG_DISCARD_DEPTHSTENCIL) != 0);
+	*ppReturnedDeviceInterface = new Direct3DDevice8(this, DeviceInterface, BehaviorFlags, PresentParams.EnableAutoDepthStencil ? PresentParams.AutoDepthStencilFormat : D3DFMT_UNKNOWN, (PresentParams.Flags & D3DPRESENTFLAG_DISCARD_DEPTHSTENCIL) != 0);
 
 	// Set default vertex declaration
 	DeviceInterface->SetFVF(D3DFVF_XYZ);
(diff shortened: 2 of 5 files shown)


WinterSnowfall:
Thank you for the explanation! Think I might have noticed it at odd times, but I doubt I'd normally classify it under "z-fighting", though at large distances I guess it may end up as such.

However, consider that D32 and D15S1 are unsupported on modern drivers and were never supported on age-accurate hardware either, based on my tests. So that leaves us with the perhaps fine difference of D16 vs D24S8/X8. Have looked at my collection of traces and D24S8/X8 indeed appears quite a bit more prolific, but D16 is not exactly what I'd call unused either.

Is the "old way" tuned for 16 bit depth?

elishacloud:
Based on my tests, the current value (old way) is too small for 16-bit z buffers, meaning that if a game chooses a different ZBias then Driect3D9 won't always honor it.  With the current code, I can create a simple test that shows objects with the same z-level appear out of order from what they should based on their ZBias.

Furthermore, the value used is far too large for 24-bit buffers.

universal963:
Unfortunately it looks like this breaks Command & Conquer Renegade again. The bullet holes on the wall flicker when moving mouse, and may even disappear at a certain angle. See the video below:

https://github.com/user-attachments/assets/f997fe31-a897-4c81-a438-c31d3008b054

elishacloud:
@universal963, can you try this build?  I think it will fix the issue: [dxwrapper.zip](https://github.com/user-attachments/files/23225236/dxwrapper.zip)

universal963:
@elishacloud This one seems to be no problem.

WinterSnowfall:
@elishacloud Dropped by to say were on the money (again :) ), because on my ancient API adventures I came across a game (Drakan), that always uses D24X8, regardless of color depth and isn't happy with D16 levels of precision. There's no need to jump through any hoops to reproduce, it's an actual ZBIAS precision bug. See the barrels in the distance in the screenshots below (their contents should not be visible):

<details>
<summary>D16</summary>

D16

</details>

<details>
<summary>D24</summary>

D24

</details>

Ended up picking between:
```
static constexpr float ZBIAS_SCALE_D16 = -1.0f / ((1 << 16) - 1); // Consider D16 precision
static constexpr float ZBIAS_SCALE_D24 = -1.0f / ((1 << 24) - 1); // Consider D24 precision
```

Depending on zbuffer bit depth, similarly to what you did, which works just fine.

elishacloud:
@WinterSnowfall, thanks for the update here.  That is the smallest precision that each buffer bit rate can handle, but I remember having an issue with them because they are smaller than the precision of a float, which is larger than that.

Maybe a better solution is:

```c++
static constexpr float ZBIAS_SCALE_D16 = -1.0f / ((1 << 16) - 1);
static constexpr float ZBIAS_SCALE_D24 = -std::max(FLT_EPSILON, 1.0f / ((1 << 24) - 1));
```

WinterSnowfall:
> @WinterSnowfall, thanks for the update here. That is the smallest precision that each buffer bit rate can handle, but I remember having an issue with them because they are smaller than the precision of a float, which is larger than that.

Oh, hmmm, interesting. Makes sense. I guess since we're "playing" in a 32-bit exclusive sandbox that will always favor FLT_EPSILON for D24. Can't say I've noticed anything off so far, but I'll make the adjustments, thanks for letting me know.

elishacloud:
I tested this with both 16-bit and 24-bit buffers.  This code worked for setting the correct z-order for different ZBias values.

```c++
static constexpr float ZBIAS_SCALE_D16 = -1.0f / ((1 << 16) - 1);
static constexpr float ZBIAS_SCALE_D24 = -FLT_EPSILON;
```

Here is the code I used calling this in EndScene().  I also tried changing the zbiasList values to all different ones to make sure that it works:

```c++
static void TestDepthBias(IDirect3DDevice9* device, DWORD depthBits)
{
	struct Vertex
	{
		float x, y, z, rhw;
		DWORD color;
	};
	DWORD FVF = (D3DFVF_XYZRHW | D3DFVF_DIFFUSE);

	// Small quad centered on screen, near depth 0.999
	float z = 0.999f;
	Vertex quads[5][4] = {
		{ {100,100,z,1,0xFFFF0000}, {200,100,z,1,0xFFFF0000}, {200,200,z,1,0xFFFF0000}, {100,200,z,1,0xFFFF0000} }, // red
		{ {120,100,z,1,0xFF00FF00}, {220,100,z,1,0xFF00FF00}, {220,200,z,1,0xFF00FF00}, {120,200,z,1,0xFF00FF00} }, // green
		{ {140,100,z,1,0xFF0000FF}, {240,100,z,1,0xFF0000FF}, {240,200,z,1,0xFF0000FF}, {140,200,z,1,0xFF0000FF} }, // blue
		{ {160,100,z,1,0xFFFFFF00}, {260,100,z,1,0xFFFFFF00}, {260,200,z,1,0xFFFFFF00}, {160,200,z,1,0xFFFFFF00} }, // yellow
		{ {180,100,z,1,0xFFFF00FF}, {280,100,z,1,0xFFFF00FF}, {280,200,z,1,0xFFFF00FF}, {180,200,z,1,0xFFFF00FF} }  // magenta
	};

	// Example test ZBias values to convert
	DWORD zbiasList[5] = { 4, 16, 0, 1, 12 };

	DWORD ZEnable = 0, ZWrite = 0, ZFunc = 0, DepthBias = 0;
	device->GetRenderState(D3DRS_ZENABLE, &ZEnable);
	device->GetRenderState(D3DRS_ZWRITEENABLE, &ZWrite);
	device->GetRenderState(D3DRS_ZFUNC, &ZFunc);
	device->GetRenderState(D3DRS_DEPTHBIAS, &DepthBias);

	device->SetFVF(FVF);
	device->SetRenderState(D3DRS_ZENABLE, TRUE);
	device->SetRenderState(D3DRS_ZWRITEENABLE, TRUE);
	device->SetRenderState(D3DRS_ZFUNC, D3DCMP_LESSEQUAL);

	for (int i = 0; i < 5; i++)
	{
		DWORD bias = GetDepthBias(zbiasList[i], depthBits);

		device->SetRenderState(D3DRS_DEPTHBIAS, bias);
		device->DrawPrimitiveUP(D3DPT_TRIANGLEFAN, 2, quads[i], sizeof(Vertex));

		// Optionally print to console
		bias = GetDepthBias(zbiasList[i], depthBits);
		Logging::Log() << __FUNCTION__ << " Bits: " << depthBits << " ZBias = " << zbiasList[i] << " -> DepthBias = " << *(float*)&bias;
	}

	device->SetRenderState(D3DRS_ZENABLE, ZEnable);
	device->SetRenderState(D3DRS_ZWRITEENABLE, ZWrite);
	device->SetRenderState(D3DRS_ZFUNC, ZFunc);
	device->SetRenderState(D3DRS_DEPTHBIAS, DepthBias);
}
```

WinterSnowfall:
@elishacloud Sorry for spotting this so late, but it's been a while since I last compiled d3d8to9. With mingw, I get the following warning(s):

> /home/builder/source/d3d8to9/source/d3d8types.cpp: In function ‘DWORD GetDepthStencilBitCount(D3DFORMAT)’:
> /home/builder/source/d3d8to9/source/d3d8types.cpp:30:16: warning: enumeration value ‘D3DFMT_UNKNOWN’ not handled in switch [-Wswitch]
>    30 |         switch (Format)
>       |                ^
> /home/builder/source/d3d8to9/source/d3d8types.cpp:30:16: warning: enumeration value ‘D3DFMT_R8G8B8’ not handled in switch [-Wswitch]
> /home/builder/source/d3d8to9/source/d3d8types.cpp:30:16: warning: enumeration value ‘D3DFMT_A8R8G8B8’ not handled in switch [-Wswitch]
> /home/builder/source/d3d8to9/source/d3d8types.cpp:30:16: warning: enumeration value ‘D3DFMT_X8R8G8B8’ not handled in switch [-Wswitch]
> /home/builder/source/d3d8to9/source/d3d8types.cpp:30:16: warning: enumeration value ‘D3DFMT_R5G6B5’ not handled in switch [-Wswitch]
> [...]

You might want to include a `default: return 0` in there to suppress them, though I know it makes no difference in practice.

crosire:
Fixed now.

PR #211, opened 2025-10-30
Fix epsilon value for DepthBias

### LINKED ISSUE #268: [Freelancer (2003)] Odd-looking shadow appears when camera moves up/down (D3d8to9 only)
When opening the equipment dealer menu in the video game Freelancer (2003), the camera moves up before the interface appears. Whenever this is done with the `D3d8to9` option enabled, an odd-looking shadow appears below the ship. If `D3d8to9` is turned off, this doesn't occur.

I've tested this on DxWrapper versions 1.1.6900.22 and 1.1.6972.22; the problem could be seen on both.

`D3d8to9 = 1`:

https://github.com/elishacloud/dxwrapper/assets/45081813/9e46e579-6993-4a4e-8640-b27265a9ae85

`D3d8to9 = 0`:

https://github.com/elishacloud/dxwrapper/assets/45081813/0e1ecb4b-2010-4d19-acd9-cd4c6587a557

elishacloud:
This fixes PR #209.  It also fixes [dxwrapper/#268](https://github.com/elishacloud/dxwrapper/issues/268)

There were two issues with the previous PR:

1. The previous PR would return 0.0f for DepthBias rather than -0.0f.

The default DepthBias is 0.0f and the default ZBias is 0.  The original code (before either PR) would multiple ZBias times -0.000005.  If ZBias was 0 then it would set DepthBias to -0.0f.  So, if a game doesn't set ZBias then DepthBias would be 0.0f.  However, if the game sets ZBias to 0 (the default) then it would actually set DepthBias to -0.0f, which is inconsistent.  Meaning, the default DepthBias would change from 0.0f to -0.0f depending on whether the game set ZBias or not.

My previous PR fixed this by forcing the DepthBias to always be 0.0f whenever a game set Zbias to 0.  However, this broke at least one game (Freelancer).  Apparently, Direct3D9 treats 0.0f and -0.0f differently.  So now this will always set DepthBias to -0.0f and sets the default state to -0.0f to keep it consistent.

2. Sets the correct epsilon number for DepthBias.

In my last PR I had the understanding that the smaller the epsilon for DepthBias, the better.  However, this didn't hold true for a number of games.  The fact is that d3d8 has a larger effect on the z-level than I understood.  With this update I did a bunch of tests to try and make DepthBias work as close to d3d8's ZBias as I could.  This update should be much closer to d3d8 than -0.000005 or than the tiny numbers I used in the last PR were.

### DIFF
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -28,8 +28,8 @@ Direct3DDevice8::Direct3DDevice8(Direct3D8 *d3d, IDirect3DDevice9 *ProxyInterfac
 	ProxyInterface->SetRenderState(D3DRS_POINTSIZE_MIN, (DWORD) 0.0f);
 	// The DEPTHBIAS value of -0.0f works differently than 0.0f
 	// Some games require defaulting to -0.0f to work correctly
-	float Value = -0.0f;
-	ProxyInterface->SetRenderState(D3DRS_POINTSIZE_MIN, *(DWORD*)&Value);
+	float DepthBias = -0.0f;
+	ProxyInterface->SetRenderState(D3DRS_DEPTHBIAS, *(DWORD*)&DepthBias);
 }
 Direct3DDevice8::~Direct3DDevice8()
 {
@@ -218,8 +218,8 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice8::Reset(D3DPRESENT_PARAMETERS8 *pPresen
 		ProxyInterface->SetRenderState(D3DRS_POINTSIZE_MIN, (DWORD) 0.0f);
 		// The DEPTHBIAS value of -0.0f works differently than 0.0f
 		// Some games require defaulting to -0.0f to work correctly
-		float Value = -0.0f;
-		ProxyInterface->SetRenderState(D3DRS_POINTSIZE_MIN, *(DWORD*)&Value);
+		float DepthBias = -0.0f;
+		ProxyInterface->SetRenderState(D3DRS_DEPTHBIAS, *(DWORD*)&DepthBias);
 	}
 
 	return hr;
--- a/source/d3d8to9_device.cpp
+++ b/source/d3d8to9_device.cpp
@@ -26,6 +26,10 @@ Direct3DDevice8::Direct3DDevice8(Direct3D8 *d3d, IDirect3DDevice9 *ProxyInterfac
 	// The default value of D3DRS_POINTSIZE_MIN is 0.0f in D3D8,
 	// whereas in D3D9 it is 1.0f, so adjust it as needed
 	ProxyInterface->SetRenderState(D3DRS_POINTSIZE_MIN, (DWORD) 0.0f);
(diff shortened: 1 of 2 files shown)


WinterSnowfall:
@elishacloud Just wanted to let you know that I've been experimenting with: `-10.0f / (1 << 16)` (~ -0.00015) in dxvk since our previous exchange on the topic and it has worked fine so far. Not sure if there's much value added in adjusting it based on depth formats as things stand, to be honest, but I'm sure it won't hurt either. The Z-fighting was indeed most likely caused by too small an epsilon, since -0.000005f is well below the threshold of precision for D16.

elishacloud:
> Just wanted to let you know that I've been experimenting with: `-10.0f / (1 << 16)` (~ -0.00015) in dxvk since our previous exchange on the topic and it has worked fine so far.

Interesting that you came up with a similar number to what I did.  Maybe I will use yours since it seems a bit better.  I was looking for something like that.

> Not sure if there's much value added in adjusting it based on depth formats as things stand, to be honest, but I'm sure it won't hurt either.

I did some tests with Freelancer and a 16bit zbuffer with a ZBias of 1 has greater effect in native d3d8 than a 24bit zbuffer with a ZBias of 1.  So as far as I can tell d3d8 does have some kind of scaling up based on the bit count of the buffer.  I don't know if any game requires this but in my tests I can create a case where d3d8to9 works differently than native d3d8 if I don't scale it up.  And, since I already have the code checked in, it's probably best to keep it.

Note: Freelancer seems to be very sensitive on this in a few locations with the shadows.  I changed the ZBias value that game and I changed the zbuffer and tested with native d3d8 and then with d3d8to9 until they both performed virtually identically.

> The Z-fighting was indeed most likely caused by too small an epsilon, since -0.000005f is well below the threshold of precision for D16.

Yes, that was definitely a problem.

WinterSnowfall:
> @WinterSnowfall, if you want a real test you can do the following:

No need, I entirely believe you if these are your findings. My point was that I played through the entirety of Freelancer both on d3d8to9 + dxvk in the past (before it supported d3d8) and on dxvk with d3d8 support and never had any issues in practice. In both cases the older epsilon of `-0.000005` was used (but I assume the game defaults to D24X8/S8, so it was never an issue).

P.S.: I'll fly over to Freeport 9 and have a closer look one of these days ;).

elishacloud:
> I played through the entirety of Freelancer both on d3d8to9 + dxvk in the past (before it supported d3d8) and on dxvk with d3d8 support and never had any issues in practice.

Yes, it is a very subtle issue.  I probably wouldn't have noticed it.  But you can see the shadow issue in the very first comment on [this thread](https://github.com/elishacloud/dxwrapper/issues/268#issue-2259516566).  I had to watch both videos several times before I saw it.  As far as I know the issue exists on all prior builds of d3d8to9 before this PR.

> P.S.: I'll fly over to Freeport 9 and have a closer look one of these days ;).

Since you are using `-10.0f / (1 << 16)` for dxvk the issue should be fixed now.  👍

PR #214, opened 2026-01-06
Use smallest epsilon number for DepthBias

elishacloud:
Sorry for the third PR on this issue!  I believe this should be the last one.

This pull request is based on the comments in PR #209.  I have tested with several games, including:  Arx Fatalis, Freelancer, Silent Hill 2.  There is no single value for the DepthBias that will work perfectly for all games.

In DX8 and older there are only 17 bias levels (0 to 16).  But there is no generic way to convert them to DX9's DepthBias.  Some games (like Freelancer) require larger DepthBias to avoid incorrect z-ordering, while other games (like Arx Fatalis and Silent Hill 2) require smaller DepthBias.

I think the best solution is use the lowest epsilon value for each buffer and then create an option for multiplying this by some other value as needed.

This PR will set it to a low epsilon value.  Then I will add an option to dxwrapper to allow it to be customized beyond that.  Since we want this d3d8to9 to work without options, I think this is the best solution for d3d8to9.

### DIFF
--- a/source/d3d8types.cpp
+++ b/source/d3d8types.cpp
@@ -51,7 +51,8 @@ DWORD CalcDepthBias(DWORD ZBias, DWORD DepthBitCount)
 	{
 	case 32:
 	case 24:
-		DepthEpsilon = -FLT_EPSILON;
+		// Bit shifting by 24 is too small for float precision. A shift of 20 seems to work best.
+		DepthEpsilon = -1.0f / ((1 << 20) - 1);
 		break;
 	default:
 	case 16:
--- a/source/d3d8types.cpp
+++ b/source/d3d8types.cpp
@@ -51,14 +51,14 @@ DWORD CalcDepthBias(DWORD ZBias, DWORD DepthBitCount)
 	{
 	case 32:
 	case 24:
-		DepthEpsilon = -10.0f / (1 << 16);
+		DepthEpsilon = -FLT_EPSILON;
 		break;
 	default:
 	case 16:
-		DepthEpsilon = -20.0f / (1 << 16);
+		DepthEpsilon = -1.0f / ((1 << 16) - 1);
 		break;
 	case 15:
-		DepthEpsilon = -25.0f / (1 << 16);
+		DepthEpsilon = -1.0f / ((1 << 15) - 1);
 		break;
 	}
 	float DepthBias = std::min(ZBias, 16UL) * DepthEpsilon;


WinterSnowfall:
I agree this is somewhat annoying to deal with, as it appears some games have different needs than others. FWIW, I ended up going with `-1.0f / ((1 << 16) - 1);` universally, which appears to work fine, as in it's still precise enough for games that use D24 and just above the threshold for those using D16. It remains to be seen if everything is content with this status quo, but for what it's worth DXVK was also using `0.000005` previously and it was mostly fine, so a value in that ballpark makes sense.

elishacloud:
Ok, after spending way too much time here I figure out that I can use a specific epsilon for 24-bit buffer and it works in every game I tested.  Using the `-1.0f / ((1 << 16) - 1);` breaks Freelancer (which uses a 24-bit buffer).  But using `-1.0f / ((1 << 20) - 1);` works for all games I tested with a 24-bit buffer.  Of course this is too small for 16-bit buffers so I will keep the `-1.0f / ((1 << 16) - 1);` for 16-bit buffers.

The latest push fixes issues with all known games I could find!

WinterSnowfall:
I can't say I see any odd things in Freelancer. The shadows cast by ships appear to be slightly above ground, which I can't remember if it happened on native or not, but I think that's a very minor problem. [That's what I see](https://drive.proton.me/urls/J1AFS57HP8#D0CZZ17rZCWq) with the "D16" compatible value above.

P.S.: If I switch to using `-1.0f / ((1 << 20) - 1)`, the shadows in that scene... disappear completely :sweat_smile:. I'm guessing Vulkan rules are a bit different than what you are seeing, so please keep doing what you see works best. I'm just fascinated and slightly annoyed with the subject, as it's a tough thing to get right, as you've noticed yourself.

elishacloud:
> P.S.: If I switch to using `-1.0f / ((1 << 20) - 1)`, the shadows in that scene... disappear completely 😅.

Yes, you are right.  The shadows are gone completely.  That is why they aren't disappearing when going up and down.  LOL.  I was just trying to notice oddity of the appearing and disappearing of the shadows when moving up and down, which happens with `-1.0f / ((1 << 16) - 1)`:

https://github.com/elishacloud/dxwrapper/assets/45081813/9e46e579-6993-4a4e-8640-b27265a9ae85

For Freelancer, I will just need to have a configurable option to allow multiplying the DepthBias.  In the case of Freelancer if you are using `-1.0f / ((1 << 20) - 1)` for 24-bit depth then you can just multiply the DepthBias by 160 to get Freelancer to work correctly, showing the shadow and not having the shadow appear and disappear when the camera is moving

If using `-FLT_EPSILON` then there are issues with games like Silent Hill 2:

https://github.com/user-attachments/assets/85ad6af1-4917-494b-aa64-04beed04fad6

**Bottom line:** I still think `-1.0f / ((1 << 20) - 1)` is better than either `-1.0f / ((1 << 16) - 1)` or `-FLT_EPSILON`.  The former is too large and the latter is too small.  `-1.0f / ((1 << 20) - 1)` still give the least artifacts in all the games I tested.

elishacloud:
@crosire, this PR is complete.  If you are ok with my tweaks in this PR you can merge the request whenever you have time.  No value is perfect, but I think this is the best compromise.  I am using the same values for dxwrapper when converting games from DX7 to DX9.

crosire:
Sounds good!

elishacloud:
Thanks @crosire!

